PR #22, opened 2022-06-22
tooling: updating the Patch version of the Git Tag to be prefixed by a `1`

tombuildsstuff:
This fixes an issue where the Go Module Proxy doesn't like Minor or Patch versions to start with a `0`, returning:

> bad request: version "v0.20220622.050833" is not canonical (wanted "")

Whilst not ideal, a `1` is at least consistent (rather than switching to use 12h time, where these would be of differing lengths)


jackofallops (approved):
LGTM 👍

PR #122, opened 2022-08-23
New Base Layer to replace go-autorest

manicminer:
Related: https://github.com/hashicorp/pandora/pull/1309


wuxu92:
Is there a schedule for this PR release? so we can use the new base layer for Automation services, I am trying to fix the broken resources that exist in azurerm. both of them are because of the autorest bug and the service team's unwilling to fix the swagger 😫

tombuildsstuff:
The Base Layer needs to account for an error being returned via a 200 OK too, see: https://github.com/hashicorp/terraform-provider-azurerm/pull/19849#discussion_r1061670890

sdk/auth/github_oidc_authorizer.go
```diff
@@ -8,23 +8,37 @@ import (
...
+	// ClientId is the client ID used when authenticating
+	ClientId string
+
+	// Environment is the Azure environment/cloud being targeted
+	Environment environments.Environment
+
+	// TenantId is the tenant to authenticate against
+	TenantId string
+
+	// AuxTenantIds lists additional tenants to authenticate against, currently only
```
tombuildsstuff, line 31:
```suggestion
	// AuxiliaryTenantIds lists additional tenants to authenticate against, currently only
```

sdk/auth/oidc_authorizer.go
```diff
@@ -8,13 +8,24 @@ import (
...
-	Environment        environments.Environment
-	Api                environments.Api
-	TenantId           string
+	// Api describes the Azure API being used
+	Api environments.Api
+
+	// TenantId is the tenant to authenticate against
+	TenantId string
+
+	// AuxTenantIds lists additional tenants to authenticate against, currently only
```
tombuildsstuff, line 20:
```suggestion
	// AuxiliaryTenantIds lists additional tenants to authenticate against, currently only
```

tombuildsstuff (approved):
Left a couple of comments inline but otherwise 👍 nice work

sdk/environments/application_ids.go
```diff
@@ -0,0 +1,115 @@
...
+	"TargetedMessagingService":          targetedMessagingServiceAppId,
+	"TeamsServices":                     teamsServiceAppId,
+	"ThreatProtection":                  threatProtectionAppId,
+	"TrafficManager":                    trafficManagerAppId,
+	"UniversalPrint":                    universalPrintAppId,
+	"WindowsDefenderAtp":                windowsDefenderAtpAppId,
+	"WindowsVirtualDesktop":             windowsVirtualDesktopAppId,
+	"Yammer":                            yammerAppId,
+
+	// TODO: do we still need these?
```
manicminer, line 112: I'd suggest yes, at least until they are fully EOL'd everywhere?

```suggestion
	// Deprecated, remove these when APIs reach final EOL
```

PR #180, opened 2022-10-13
Add Microsoft-SecurityEvent support

olafhartong:
The data collection endpoint Microsoft-SecurityEvent is missing from this module. 
This is a required endpoint for the Azure Sentinel connector which makes use of the Azure Monitor Agent.

The added code has been tested and is working to deploy our labs.

### DIFF
--- a/resource-manager/insights/2021-04-01/datacollectionrules/constants.go
+++ b/resource-manager/insights/2021-04-01/datacollectionrules/constants.go
@@ -78,6 +78,7 @@ const (
 	KnownDataFlowStreamsMicrosoftNegativePerf            KnownDataFlowStreams = "Microsoft-Perf"
 	KnownDataFlowStreamsMicrosoftNegativeSyslog          KnownDataFlowStreams = "Microsoft-Syslog"
 	KnownDataFlowStreamsMicrosoftNegativeWindowsEvent    KnownDataFlowStreams = "Microsoft-WindowsEvent"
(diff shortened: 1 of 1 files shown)


olafhartong:
To add some addtional context, the terraform-provider-azurerm plugin relies on this sdk. The feature that requires this modification is azurerm_monitor_data_collection_rule.

With the current build the Microsoft-SecurityEvent stream is not permitted, as the sdk, where the plugin relies on does not support it.

jackofallops:
Hi @olafhartong  - Thanks for this PR. The SDK's here are generated, and not manually updated, so I'm afraid we cannot accept this change as it will simply be overwritten next time the files are generated.  The generator for this process is the recently open sourced [Pandora](https://github.com/hashicorp/pandora) project. 
If items are missing from the SDK, they're likely missing from the Azure API Specs, meaning that the source data is the place to fix this unfortunately.  In this instance, the enum you are attempting to add is missing from [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/monitor/resource-manager/Microsoft.Insights/stable/2021-04-01/dataCollectionRules_API.json#L472) I think? If you can submit an issue / pr there to fix the spec appropriately? Once the source data is correct, the automation tasks to ingest the Swagger will pull through the missing enum and will generate the same for the SDK here.

Thanks again!

olafhartong:
Ah cool. I didn’t know that. Will do a PR there, thanks for the quick response.

jackofallops:
> Ah cool. I didn’t know that. Will do a PR there, thanks for the quick response.

Thanks for understanding, and for the contribution!  We'll take a look at making this info more obvious in the project README 👍

PR #264, opened 2023-01-24
github: removing the spurious `read` permission - the manually specified GHA token should be sufficient

tombuildsstuff:


jackofallops (approved):
LGTM 👍

PR #282, opened 2023-02-04
Auto PR: Regenerating the Go SDK (3baf99283f924a71225066d7957b1ced563f7733)

hc-github-team-tf-azure:
Regenerating the SDK based on the latest changes


manicminer (approved):

tombuildsstuff:
@manicminer thinking about this - rolling this out is going to mean we're (temporarily) blocking upgrading the SDK in the provider until everything's in, since each of these 3 packages are currently imported - would it be worth picking a new service (e.g. chaos)  instead which is imported but not used in the provider right now to enable us to get these changes in, without blocking new releases?

manicminer:
@tombuildsstuff No problem I can pare this back a little. I updated the corresponding services already in #20320 but it makes sense to add these later so there's ample time to review the other changes in isolation.

PR #283, opened 2023-02-05
Autorest compatibility layer for the auth package

manicminer:


sdk/auth/autorest/auth.go
```diff
@@ -0,0 +1,93 @@
...
+				}
+
+				auxTokenList := make([]string, 0)
+				for _, a := range auxTokens {
+					if a != nil && a.AccessToken != "" {
+						auxTokenList = append(auxTokenList, fmt.Sprintf("%s %s", a.TokenType, a.AccessToken))
+					}
+				}
+
+				return autorest.Prepare(req, autorest.WithHeader("x-ms-authorization-auxiliary", strings.Join(auxTokenList, ", ")))
```
tombuildsstuff, line 52: does calling this with an empty slice/value cause issues? I'd have thought we shouldn't be setting this unless there's multiple tokens present?

manicminer: This seems to be fine with an empty slice... but we get an empty header which seems messy; will fix 👍

tombuildsstuff (approved):

PR #298, opened 2023-02-13
`sdk/client`: update `(*Response).Unmarshal` for `[]byte` type

wuxu92:
And update the Test code for response.Unmarshal for different types.

This will workaround https://github.com/hashicorp/pandora/pull/1756 when working with the new Base layer.

```
=== RUN   TestAccClient
--- PASS: TestAccClient (1.36s)
PASS
```

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -218,14 +218,12 @@ func (r *Response) Unmarshal(model interface{}) error {
 
 		// copy the byte stream across
 		switch bs := model.(type) {
-		case []byte:
-			model = respBody
 		case *[]byte:
 			*bs = respBody
 		case **[]byte:
 			*bs = &respBody
 		default:
-			return fmt.Errorf("internal-error: `model` must be []byte but got %+v", model)
+			return fmt.Errorf("internal-error: `model` must be *[]byte or **[]byte but got %+v", model)
 		}
 	}
 
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -161,7 +161,8 @@ func (r *Response) Unmarshal(model interface{}) error {
 	isContentType := func(typ string) bool {
 		return strings.Contains(contentType, typ)
 	}
-	if isContentType("application/json") {
+	switch {
+	case isContentType("application/json"):
 		// Read the response body and close it
 		respBody, err := io.ReadAll(r.Body)
 		if err != nil {
@@ -179,9 +180,7 @@ func (r *Response) Unmarshal(model interface{}) error {
 
 		// Reassign the response body as downstream code may expect it
 		r.Body = io.NopCloser(bytes.NewBuffer(respBody))
-	}
-
-	if isContentType("application/xml") || isContentType("text/xml") {
+	case isContentType("application/xml") || isContentType("text/xml"):
 		// Read the response body and close it
 		respBody, err := io.ReadAll(r.Body)
 		if err != nil {
@@ -199,9 +198,7 @@ func (r *Response) Unmarshal(model interface{}) error {
 
 		// Reassign the response body as downstream code may expect it
 		r.Body = io.NopCloser(bytes.NewBuffer(respBody))
-	}
-
-	if isContentType("application/octet-stream") || strings.HasPrefix(contentType, "text/") {
+	case isContentType("application/octet-stream") || strings.HasPrefix(contentType, "text/"):
 		// Read the response body and close it
 		respBody, err := io.ReadAll(r.Body)
 		if err != nil {
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -69,6 +69,12 @@ func RetryableErrorHandler(resp *http.Response, _ error, _ int) (*http.Response,
 	return resp, nil
 }
 
+type contentType string
+
+func (c contentType) is(typ string) bool {
+	return strings.Contains(string(c), typ)
+}
+
 // Request embeds *http.Request and adds useful metadata
 type Request struct {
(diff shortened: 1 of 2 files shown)


manicminer:
Thanks for this suggestion @wuxu92. What does a `text/powershell` response look like? Is this a unicode format? Can we parse it?

wuxu92:
it's form automation spec:  https://github.com/Azure/azure-rest-api-specs/blob/3034ada77d465b317cda51250a92454824cca06b/specification/automation/resource-manager/Microsoft.Automation/stable/2019-06-01/dscConfiguration.json#L281-L283 

if I remember correctly, this response `text/powershell` can be treated as literal string.

wuxu92:
that detect plain text response as a string would be a better choice

tombuildsstuff:
Yeah, auto-detecting from a content-type prefix of `text/` seems reasonable here, the challenge is should we expose that as a string (and assume unicode etc), or to expose that as a set of bytes? I think the output format (e.g. []byte/string) is going to need to change depending on what's being returned though, I could see PowerShell making sense as a byte array here?

wuxu92:
it seems the sdk generator will use `*[]byte` for all `Model` field if http response is not `json`. and treate `text/poweshell`  as []ybte make sense to me.

wuxu92:
@manicminer @tombuildsstuff how do you think the current version of response.Marshal? I have updated this PR to treat all `text/xxx` content-type as text and try to copy the body to `[]byte` or `*string` types.

sdk/client/client.go
```diff
@@ -196,13 +198,7 @@ func (r *Response) Unmarshal(model interface{}) error {
...
 		// Reassign the response body as downstream code may expect it
 		r.Body = io.NopCloser(bytes.NewBuffer(respBody))
-	}
-
-	if strings.Contains(contentType, "application/octet-stream") {
-		if _, ok := model.([]byte); !ok {
-			return fmt.Errorf("internal-error: `model` must be []byte but got %+v", model)
-		}
-
+	case isContentType("application/octet-stream") || strings.HasPrefix(contentType, "text/"):
```
tombuildsstuff, line 201: We parse this from the Swagger into a `RawFile` object: https://github.com/hashicorp/pandora/blob/a912fd797a6f61a032649cb4239cf716c442e4a6/data/Pandora.Definitions.ResourceManager/Automation/v2022_08_08/DscConfiguration/Operation-GetContent.cs#L16-L27

Which [when output in the Go SDK](https://github.com/hashicorp/pandora/blob/a912fd797a6f61a032649cb4239cf716c442e4a6/tools/sdk/resourcemanager/models.go#L83-L84) is output as `[]byte` - so I think we can simplify this to:

```go
if strings.Contains(contentType, "application/octet-stream") || strings.Contains(contentType, "text/powershell") {
```

and then re-use the `application/octet-stream` logic? Should we need to parse additional types in the future we can handle that, but we _shouldn't_ be doing so since the Importer will normalize this for us - so we can handle just the simple case for now

---

We'll also need to add similar support to the Marshal function, since the `ReplaceContent` operation also sends a `RawFile`: https://github.com/hashicorp/pandora/issues/2185

tombuildsstuff (changes requested):
hey @wuxu92 

Apologies for the delayed review here, digging into this whilst we previously suggested trying to handle all `text/` types, digging through the source data it appears this is probably going to be more practical to limit this to just `text/powershell` for now, meaning that we can assume a `RawFile` (which is output as a `[]byte`) is used.

A `RawFile` is intentionally output as a byte array rather than a string since the encoding could differ - so it seems worthwhile only supporting `[]byte` at this time until we have further information on other payloads.

Whilst I appreciate that's different from the suggestion above, we're also relying on being able to correctly marshal this data, as detailed in https://github.com/hashicorp/pandora/issues/2185 - so we'll need to update the `Marshal` function here too - would you mind taking a look?

Thanks!

sdk/client/client.go
wuxu92: thanks! I have jsut upadted this PR. as i found the generated response Model field is a `*[]byte` 
https://github.com/hashicorp/go-azure-sdk/blob/bd4de95c1ffb70323a9dec602b723a3bf0dc908f/resource-manager/automation/2022-08-08/dscconfiguration/method_getcontent_autorest.go#L15-L18

and it pass a `&result.Model` (is a `**[]byte` now) to Marshal. so i handled both `*[]byte` and `**[]byte` in this PR.
https://github.com/hashicorp/go-azure-sdk/blob/bd4de95c1ffb70323a9dec602b723a3bf0dc908f/resource-manager/automation/2022-08-08/dscconfiguration/method_getcontent_autorest.go#L64

wuxu92:
@tombuildsstuff i have updated the PR, would you please have a look again?

tombuildsstuff:
hey @wuxu92 

Thanks for pushing those changes to this PR.

Whilst this is moving closer towards where this needs to be for us to get this merged, chatting internally we want to take a slightly different direction for this one by adding unit tests covering both the Marshal and Unmarshal methods since both have a number of expectations.

The end result of this is that we should be able to simplify this change, and now that https://github.com/hashicorp/pandora/pull/2332 has been merged, it should be possible to add support for this functionality.

To that end, after chatting internally about this I've opened #445 - which both makes these changes, and then adds support for both Marshaling and Unmarshaling `*[]byte` values, including PowerShell script.

I've validated this change in the Provider, as such whilst I'd like to thank you for this contribution - I'm going to close this PR in favour of #445 which'll (once threaded through) will enable fixing this issue.

Thanks!

PR #300, opened 2023-02-13
Environments: Add app IDs for some Office and Azure Postgres applications

manicminer:
Related: https://github.com/manicminer/hamilton/pull/216


sdk/environments/constants.go
```diff
@@ -71,10 +75,14 @@ const office365InformationProtectionAppId = "2f3f02c9-5679-4a5c-a605-0de55b07d13
...
 const office365PortalAppId = "00000006-0000-0ff1-ce00-000000000000"
 const office365SharePointOnlineAppId = "00000003-0000-0ff1-ce00-000000000000"
+const office365SuiteUxAppId = "4345a7b9-9a63-4910-a426-35363201d503"
 const office365ZoomAppId = "0d38933a-0bbd-41ca-9ebd-28c4b5ba7cb7"
+const officeHomeAppId = "4765445b-32c6-49b0-83e6-1d93765276ca"
+const officeUwpPwaAppId = "0ec893e0-5785-4de6-99da-4ed124e5296c"
 const oneNoteAppId = "2d4d3d8e-2be3-4bef-9f87-7875a61c29de"
 const oneProfileServiceAppId = "b2cc270f-563e-4d8a-af47-f00963a71dcd"
 const ossRDBMSAppId = "123cd850-d9df-40bd-94d5-c9f07b7fa203"
+const ossRdbmsPostgreSqlFlexibleServerAadAuthenticationAppId = "5657e26c-cc92-45d9-bc47-9da6cfdb4ed9"
```
tombuildsstuff, line 85: we should also add these to the Base Environment?

tombuildsstuff (changes requested):
All the other items are present in the `base` environment, so these are available as typed objects against the Environment - so we should replicate that for these?

sdk/environments/constants.go
manicminer: Yes indeed, I missed that! 😬

manicminer: I've added these to the base environment, and also updated the `Api` interface - the `MicrosoftGraphAppId()` method is now renamed to `AppId()` since it seems we missed that in the initial PR. We don't have any usages of that just yet.

tombuildsstuff (approved):

PR #312, opened 2023-02-16
Test fix and OData tweaks

manicminer:
- Metadata client tests to run only as needed, similar to auth tests
- Refactor OData and Client types for easier re-use

### DIFF
--- a/sdk/client/options.go
+++ b/sdk/client/options.go
@@ -13,35 +13,71 @@ type Options interface {
 	ToQuery() *QueryParams
 }
 
+// Headers is a representation of the HTTP headers to be sent with a Request
 type Headers struct {
-	val map[string]string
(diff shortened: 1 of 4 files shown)


sdk/client/options.go
```diff
@@ -13,42 +13,92 @@ type Options interface {
...
+	}
+	if q2.vals == nil {
+		return
+	}
+	for k, v := range q2.vals {
+		q.vals[k] = v
+	}
+}
+
+// Values returns a url.Values mao containing query parameter values
```
tombuildsstuff, line 101:
```suggestion
// Values returns a url.Values map containing query parameter values
```

tombuildsstuff (approved):
LGTM 👍

PR #318, opened 2023-02-17
sdk: fixing a bug where the authorizer would enter a crash loop when using MSI

tombuildsstuff:


manicminer (approved):

tombuildsstuff:
Fixes the MSI bug: https://github.com/hashicorp/terraform-provider-azurerm/issues/20514#issuecomment-1434444308

PR #327, opened 2023-02-21
Switch PKCS#12 library to software.sslmate.com/src/go-pkcs12

manicminer:
golang.org/x/crypto/pkcs12 is unmaintained, this replacement is license compatible and more up-to-date.

This wins us the following:
- Support for parsing out certificate chains stored in PKCS#12 bundles
- Support for SHA256 HMAC which is the default hash function for PKCS#12 bundles with newer versions of OpenSSL/LibreSSL


manicminer:
closing, actions are broken

PR #339, opened 2023-02-28
SDK: updating the  pollers for `resourcemanager`

tombuildsstuff:
This PR fixes a number of issues identified when enabling the base layer in https://github.com/hashicorp/terraform-provider-azurerm/pull/20688, namely:

1. Some Resource Manager APIs return the LRO OperationStatus URI on a verbose hostname (including the port), such that the check we have present here isn't strictly needed (else we fail comparing `https://management.azure.com` to `https://management.azure.com:443`, which whilst technically the same the LRO URI isn't guaranteed to be on the same port afaik)
2. Some Resource Manager APIs (such as SignalR) return an LRO Status of `Running` rather than `InProgress`, as such this PR adds an exception for this
3. Introduces a Delete LRO - which (after deleting a resource) will validate that the resource ultimately disappears


jackofallops (approved):
LGTM 👍

PR #342, opened 2023-03-01
Add OData types of Windows Autopilot

kenchan0130:
We always need the value of the OData type in order to call the API of Windows Autopilot.
I have added OData types related to Windows Autopilot.

* https://learn.microsoft.com/en-us/graph/api/intune-enrollment-activedirectorywindowsautopilotdeploymentprofile-create?view=graph-rest-beta
* https://learn.microsoft.com/en-us/graph/api/intune-enrollment-azureadwindowsautopilotdeploymentprofile-get?view=graph-rest-beta

Related to https://github.com/manicminer/hamilton/pull/228


manicminer (approved):
Thanks @kenchan0130, LGTM 👍

PR #353, opened 2023-03-07
resourcemanager/poller: splitting Delete and non-Delete LROs

tombuildsstuff:
Accounting for some behavioural issues spotted in https://github.com/hashicorp/terraform-provider-azurerm/pull/20810


jackofallops (approved):
LGTM 👍

PR #362, opened 2023-03-09
auth: Use a custom function to determine access token expiry

manicminer:
The default method `*oauth2.Token{}.Valid()` has a hardcoded delta of 10 seconds, which means that tokens are only renewed 10 seconds before they expire. This leads to race conditions during long running operations with the Resource Manager API, when a polling request is built immediately prior to this window, but only sent on or after the expiry time. Additionally, it's plausible that 1 (or more) second(s) may be lost during the process of token issuance, since it's likely the token is generated prior to being returned by the API, and it is the latter that informs us of the expiry time.

This change extends this window to 10 minutes for any access token with a validity period of 20+ minutes. For tokens having a validity period less than 20 minutes, those are renewed when 50% or more of that validity period has elapsed.

For example, a token that is valid for 1 hour (very common) will be renewed after it has been held for 50 minutes. However, a token issued for 10 minutes will be renewed when it has been held for 5 minutes (i.e. >=50% of the validity period).

Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/20834

### DIFF
--- a/sdk/auth/token.go
+++ b/sdk/auth/token.go
@@ -1,0 +1,44 @@
+package auth
+
+import (
+	"golang.org/x/oauth2"
+	"time"
+
+	"github.com/hashicorp/go-azure-sdk/sdk/claims"
+)
+
+const tokenExpiryDelta = 10 * time.Minute
+
+// tokenExpiresSoon returns true if the token expires within 10 minutes, or if more than 50% of its validity period has elapsed (if this can be determined), whichever is later
+func tokenDueForRenewal(token *oauth2.Token) bool {
+	if token == nil {
+		return true
+	}
+
+	// Some tokens may never expire
+	if token.Expiry.IsZero() {
+		return false
+	}
+
+	expiry := token.Expiry.Round(0)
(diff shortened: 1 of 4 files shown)


sdk/auth/token.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	// Some tokens may never expire
+	if token.Expiry.IsZero() {
+		return false
+	}
+
+	expiry := token.Expiry.Round(0)
+	delta := tokenExpiryDelta
+	now := time.Now()
```
tombuildsstuff, line 25: IIRC the access token expiration time is in UTC, so do we want to explicitly convert these to UTC too?

tombuildsstuff (approved):
One comment about timezones but otherwise this LGTM 👍

sdk/auth/token.go
manicminer: It's in local time actually, we're computing the expiration time based on the `expires_in` field from the login response, which is a seconds value that we add to the local time before storing the result.

PR #367, opened 2023-03-13
auth: use a retryable http client to handle 429 responses from token endpoint(s)

manicminer:
Docs: https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/resilience-client-app
Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/20901

### DIFF
--- a/sdk/auth/client.go
+++ b/sdk/auth/client.go
@@ -1,0 +1,14 @@
+package auth
+
+import (
+	"net/http"
+
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// httpClient returns a shimmed retryablehttp Client, at the moment with default retry,
+// backoff and timeout settings, but we may need to change this in the future.
+func httpClient() *http.Client {
+	retryableClient := retryablehttp.NewClient()
+	return retryableClient.StandardClient()
+}
--- a/sdk/auth/client.go
+++ b/sdk/auth/client.go
@@ -14,6 +14,8 @@ import (
 )
 
 type httpClientParams struct {
+	instanceMetadataService bool
+
 	retryWaitMin  time.Duration
 	retryWaitMax  time.Duration
 	retryMaxCount int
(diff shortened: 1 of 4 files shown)


sdk/auth/client.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"net/http"
+
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// httpClient returns a shimmed retryablehttp Client, at the moment with default retry,
+// backoff and timeout settings, but we may need to change this in the future.
+func httpClient() *http.Client {
+	retryableClient := retryablehttp.NewClient()
```
tombuildsstuff, line 12: double-checking this includes the ProxyFromEnvironment setting?

sdk/auth/managed_identity_authorizer.go
```diff
@@ -169,10 +169,10 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
 	req.Header = http.Header{
 		"Metadata": []string{"true"},
 	}
-	client := &http.Client{
-		Transport: http.DefaultTransport,
-		Timeout:   msiDefaultTimeout,
-	}
+
+	client := httpClient()
+	client.Timeout = msiDefaultTimeout
```
tombuildsstuff, line 174: double-checking: are there any other configuration changes we'd need to make to the retryable client given the Managed Identity API is must more limited with regards to retries than resource manager? https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance

tombuildsstuff (approved):
A couple of questions inline to sanity-check but otherwise 👍

sdk/auth/client.go
manicminer: Fortunately it does, thanks to very [usable defaults](https://github.com/hashicorp/go-cleanhttp/blob/02f12f05b908335f65f124c1a7f1ec45f5c42a35/cleanhttp.go#L26-L42) in go-cleanhttp.

sdk/auth/managed_identity_authorizer.go
manicminer: Thanks for the docs link, we should probably add a custom Backoff func for this.

tombuildsstuff (approved):
One thought but otherwise 👍

PR #371, opened 2023-03-14
`auth` - Support Retry for Managed Identity authentication

ms-zhenhua:
### Problem Description

Since [AIMS](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows) has [max request limitation](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#rate-limiting) per second, I migrate the retry logic from [autorest](https://github.com/Azure/go-autorest/blob/ee71315119d4d7088d74ca9fcbc7301ce2ed2bc1/autorest/adal/token.go#L1167), which will help resolve problems caused by this limitation.

### Steps to Reproduce
1. Create a VM with `managed identity` enabled
2. Run `terraform plan -parallelism=100` on the VM with the attached configuration(set your own `subscription_id` and `tenant_id` before running)
[err.zip](https://github.com/hashicorp/go-azure-sdk/files/10966251/err.zip)

### Panic Output
image

### DIFF
--- a/sdk/auth/managed_identity_authorizer.go
+++ b/sdk/auth/managed_identity_authorizer.go
@@ -157,7 +157,11 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
 }
 
 func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
-	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*30))
+	const retryMaxCount = 5
+	const retryWaitMax = 60
+	const singleTimeout = 10
+	const timeout = retryMaxCount * (retryWaitMax + singleTimeout)
+	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*(30+timeout)))
 	defer cancel()
 
 	var req *http.Request
@@ -173,9 +177,9 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
 		instanceMetadataService: true,
 
 		retryWaitMin:  2 * time.Second,
-		retryWaitMax:  60 * time.Second,
-		retryMaxCount: 5,
-		timeout:       10 * time.Second,
+		retryWaitMax:  retryWaitMax * time.Second,
+		retryMaxCount: retryMaxCount,
+		timeout:       timeout * time.Second,
 		useProxy:      false,
 	})
 
--- a/sdk/auth/managed_identity_authorizer.go
+++ b/sdk/auth/managed_identity_authorizer.go
@@ -157,15 +157,13 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
 }
 
 func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
-	const retryMaxCount = 5
-	const retryWaitMax = 60
-	const singleTimeout = 10
-	const timeout = retryMaxCount * (retryWaitMax + singleTimeout)
-	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*(30+timeout)))
-	defer cancel()
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		return nil, fmt.Errorf("internal-error: context had no deadline")
+	}
 
 	var req *http.Request
-	req, err = http.NewRequestWithContext(ctx2, http.MethodGet, url, http.NoBody)
+	req, err = http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
 	if err != nil {
 		return
 	}
@@ -177,9 +175,9 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
 		instanceMetadataService: true,
 
 		retryWaitMin:  2 * time.Second,
-		retryWaitMax:  retryWaitMax * time.Second,
-		retryMaxCount: retryMaxCount,
-		timeout:       timeout * time.Second,
+		retryWaitMax:  60 * time.Second,
+		retryMaxCount: 5,
+		timeout:       time.Until(deadline),
 		useProxy:      false,
 	})
 


sdk/auth/managed_identity_authorizer.go
```diff
@@ -72,7 +73,7 @@ func (a *ManagedIdentityAuthorizer) Token(ctx context.Context, _ *http.Request)
 
 	url := fmt.Sprintf("%s?%s", a.conf.MsiEndpoint, query.Encode())
 
-	body, err := azureMetadata(ctx, url)
+	body, err := retryForIMDS(ctx, url)
```
tombuildsstuff, line 76: this is already done via the `retryablehttp.Client` [which is instantiated here](https://github.com/hashicorp/go-azure-sdk/blob/4ddf8bbe31e99f92490e553935e836cf2ac81201/sdk/auth/managed_identity_authorizer.go#L172-L180) which [contains some IDMS specific logic](https://github.com/hashicorp/go-azure-sdk/blob/4ddf8bbe31e99f92490e553935e836cf2ac81201/sdk/auth/client.go#L50-L56) - so I believe this change is already present in `main` - is this still an issue when vendoring the latest `main` into the provider?

sdk/auth/managed_identity_authorizer.go
```diff
@@ -191,9 +192,77 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
...
+		http.StatusNotFound,
+		http.StatusGone,
+		// all remaining 5xx
+		http.StatusNotImplemented,
+		http.StatusHTTPVersionNotSupported,
+		http.StatusVariantAlsoNegotiates,
+		http.StatusInsufficientStorage,
+		http.StatusLoopDetected,
+		http.StatusNotExtended,
+		http.StatusNetworkAuthenticationRequired)
```
tombuildsstuff, line 224: we can remove this function - but (almost) all of these would be terminal failure modes, so shouldn't be something we should retry on

sdk/auth/managed_identity_authorizer.go
```diff
@@ -156,7 +157,7 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
 	})
 }
 
-func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
+func azureMetadata(ctx context.Context, url string) (body []byte, statusCode int, err error) {
 	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*30))
 	defer cancel()
```
tombuildsstuff, line 162: there's a legitimate issue here however, where the context deadline may be exceeded before the request has completed, should the IDMS service become unavailable (where it can have a 70 second delay) - so we should instead push this up:

```suggestion
	if _, ok := ctx.Deadline(); !ok {
		return nil, fmt.Errorf("internal-error: context had no deadline")
	}
```

which'll require calls to the token function to have a context with a deadline attached, which is the expected behaviour here here

tombuildsstuff (changes requested):
hi @ms-zhenhua 

Thanks for this PR - I've taken a look through and left some comments inline, I believe the latest changes in `main` should have resolved most of these since `hashicorp/go-retryablehttp` takes care of this for us, with #367 including the specific retry logic for IDMS - so I'm wondering if this is still needed with `main` vendored into the provider?

Thanks!

sdk/auth/managed_identity_authorizer.go
tombuildsstuff: ah, apologies this isn't in main yet: https://github.com/hashicorp/terraform-provider-azurerm/pull/20918

ms-zhenhua: Hi @tombuildsstuff, thank you for reviewing.
I found the latest main branch will still cause timeout issue. The reason should be the [timeout ](https://github.com/hashicorp/go-azure-sdk/blob/cacb11e1b5b1f87eb2809ecb9a256cb4a57187c1/sdk/auth/managed_identity_authorizer.go#L178)value is too short. This timeout value is used for all retried http requests, not one of them. After updating the timeout value and the [context deadline](https://github.com/hashicorp/go-azure-sdk/blob/cacb11e1b5b1f87eb2809ecb9a256cb4a57187c1/sdk/auth/managed_identity_authorizer.go#L160), this issue is resolved. Kindly help take another review of this PR. Thanks.

image

sdk/auth/managed_identity_authorizer.go
```diff
@@ -157,7 +157,11 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
 }
 
 func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
-	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*30))
+	const retryMaxCount = 5
+	const retryWaitMax = 60
+	const singleTimeout = 10
+	const timeout = retryMaxCount * (retryWaitMax + singleTimeout)
+	ctx2, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*(30+timeout)))
 	defer cancel()
```
tombuildsstuff, line 165: as mentioned in the last review, we shouldn't be using a timed context here (I appreciate this was already present, but should have been removed), since it hides the reliance on a timed context - we should instead be requiring that the `ctx` passed has a deadline attached:

```suggestion
	deadline, ok := ctx.Deadline()
	if !ok {
		return fmt.Errorf("internal-error: context had no deadline")
	}
```

sdk/auth/managed_identity_authorizer.go
```diff
@@ -173,9 +177,9 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
 		instanceMetadataService: true,
 
 		retryWaitMin:  2 * time.Second,
-		retryWaitMax:  60 * time.Second,
-		retryMaxCount: 5,
-		timeout:       10 * time.Second,
+		retryWaitMax:  retryWaitMax * time.Second,
+		retryMaxCount: retryMaxCount,
```
tombuildsstuff, line 181: these don't need to be configurable:

```suggestion
		retryWaitMax:  60 * time.Second,
		retryMaxCount: 5,
```

sdk/auth/managed_identity_authorizer.go
```diff
@@ -173,9 +177,9 @@ func azureMetadata(ctx context.Context, url string) (body []byte, err error) {
...
+		timeout:       timeout * time.Second,
```
tombuildsstuff, line 182: we can use the timeout on the context:

```suggestion
		timeout:       time.Until(deadline),
```

tombuildsstuff (changes requested):
hi @ms-zhenhua 

Thanks for pushing those changes, I've taken a look through and left some more comments inline, but if we can fix those up then this should be good to go 👍 

Thanks!

sdk/auth/managed_identity_authorizer.go
ms-zhenhua: I got the following error with this code. Seems the deadline does not exist at this time. I think we may still need to set a deadline here. 
image

tombuildsstuff: Yep, we'll need to update the calling code too, that's expected given the new check

manicminer:
As we are encountering test issues in the provider, I've opened https://github.com/hashicorp/go-azure-sdk/pull/375 to get this fixed quickly. Thanks for the headsup @ms-zhenhua.

Issue #381, opened 2023-03-20
Issue with Databricks / Databricks Virtual Network Peerings / API Version 2023-02-01


WodansSon:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Community Note

* Please vote on this issue by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Service Used

Databricks

### API Versions Used

2023-02-01

### Description

Request/Response Chain:

```
DELETE /subscriptions/{subscription}/resourceGroups/XXXXX-databricks-vnet-peering/providers/Microsoft.Network/virtualNetworks/remote-vnet/virtualNetworkPeerings/peer-to-databricks?api-version=2022-07-01 HTTP/1.1
Host: management.azure.com
User-Agent: Go/go1.19.6 (amd64-linux) go-autorest/v14.2.1 tombuildsstuff/kermit/v0.20230307.1105329 network/2022-07-01 HashiCorp Terraform/1.3.9 (+https://www.terraform.io) Terraform Plugin SDK/2.10.1 terraform-provider-azurerm/dev pid-222c6c49-1b0a-5959-a213-6608f9eb8820
X-Ms-Correlation-Request-Id: c60b7c62-6771-97a5-d3a2-544fce6fc059
Accept-Encoding: gzip: timestamp=2023-03-20T14:31:41.959-0600
2023-03-20T14:31:48.894-0600 [DEBUG] provider.terraform-provider-azurerm: AzureRM Response for https://management.azure.com/subscriptions/{subscription}/resourceGroups/XXXXX-databricks-vnet-peering/providers/Microsoft.Network/virtualNetworks/remote-vnet/virtualNetworkPeerings/peer-to-databricks?api-version=2022-07-01:
HTTP/2.0 202 Accepted
Azure-Asyncnotification: Enabled
Azure-Asyncoperation: https://management.azure.com/subscriptions/{subscription}/providers/Microsoft.Network/locations/westeurope/operations/fdcb4e39-d6ab-46ae-b6c0-b9886ba88b4a?api-version=2022-07-01
Cache-Control: no-cache
Date: Mon, 20 Mar 2023 20:31:49 GMT
[...]

PR #383, opened 2023-03-21
Bugfix: OData parsing for bad responses

manicminer:
Intentionally ignore errors when parsing OData for unpaged responses - this can fail for various reasons but notably when the JSON response is technically valid but comprises a non-object literal such as a string or int.

We are only parsing OData from the response here in order to pass it along to a provided ValidStatusFunc, and this is a best-effort as the provided *odata.OData may be nil.

Fixes: #381

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -400,9 +400,6 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 	if err != nil {
 		return resp, err
 	}
-	if resp == nil {
-		return resp, fmt.Errorf("nil response received")
-	}
 
 	// Determine whether response status is valid
 	if !containsStatusCode(req.ValidStatusCodes, resp.StatusCode) {
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -394,14 +394,14 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 		}
 	}
 
-	// Extract OData from response
-	resp.OData, err = odata.FromResponse(resp.Response)
-	if err != nil {
-		return resp, err
-	}
+	// Extract OData from response, intentionally ignoring any errors as it's not crucial to extract
+	// valid OData at this point (valid json can still error here, such as any non-object literal)
+	resp.OData, _ = odata.FromResponse(resp.Response)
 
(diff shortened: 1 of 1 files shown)


manicminer:
Screenshot 2023-03-21 at 10 45 47

sdk/client/client.go
```diff
@@ -394,27 +394,23 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
...
-		return resp, fmt.Errorf("nil response received")
-	}
+	// Extract OData from response, intentionally ignoring any errors as it's not crucial to extract
+	// valid OData at this point (valid json can still error here, such as any non-object literal)
+	resp.OData, _ = odata.FromResponse(resp.Response)
 
 	// Determine whether response status is valid
 	if !containsStatusCode(req.ValidStatusCodes, resp.StatusCode) {
-		if f := req.ValidStatusFunc; f != nil && f(resp.Response, o) {
+		// The status code didn't match, but we also need to check the ValidStatusFUnc, if provided
```
tombuildsstuff, line 403:
```suggestion
		// The status code didn't match, but we also need to check the ValidStatusFunc, if provided
```

sdk/client/client.go
```diff
@@ -394,27 +394,23 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
...
-	resp.OData, err = odata.FromResponse(resp.Response)
-	if err != nil {
-		return resp, err
-	}
-	if resp == nil {
-		return resp, fmt.Errorf("nil response received")
-	}
+	// Extract OData from response, intentionally ignoring any errors as it's not crucial to extract
+	// valid OData at this point (valid json can still error here, such as any non-object literal)
+	resp.OData, _ = odata.FromResponse(resp.Response)
```
tombuildsstuff, line 399: can we add a unit test to cover #381?

tombuildsstuff (approved):
LGTM - but we should add a test to cover this scenario

sdk/client/client.go
tombuildsstuff: sounds reasonable to me, I don't think it blocks this PR - but we should look to make that more testable/add a PR to add a test for that in the future 👍

PR #415, opened 2023-04-17
Bugfix: save the expiry time for access tokens from Azure CLI

manicminer:


jackofallops (approved):
Thanks @manicminer - LGTM 👍

PR #417, opened 2023-04-17
Add support for Delta Token

MarkDordoy:
Adding basic support for a DeltaToken odata value to be passed. 

Having this will allow Hamilton and other consuming libraries to pass a deltaToken when using delta based endpoints.


manicminer (approved):
Many thanks @MarkDordoy, this LGTM

PR #419, opened 2023-04-19
Enable logger for the request retryable client - same as `auth.Client`

magodo:
Same as #382, but for the retryable client initialized during executing a `Request`. This and the `auth.Client` are the only two places that initialize the retryable client, hence needing this logger setting.


tombuildsstuff (approved):
LGTM - thanks for this @magodo

PR #425, opened 2023-04-21
Ensure POST with not req body also defined the header `Content-Length`

magodo:
Ensure POST with not req body also defined the header `Content-Length`. The motivation of this PR is because I've encountered error from Azure service complaining the POST request has not defined the `Content-Lenght` in header:

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Length Required</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
<BODY><h2>Length Required</h2>
<hr><p>HTTP Error 411. The request must be chunked or have a content length.</p>
</BODY></HTML>
```

See: https://github.com/hashicorp/terraform-provider-azurerm/pull/21481 for details.

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -316,6 +316,12 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 
 	var err error
 
+	// Ensure the Content-Lenght header is set for methods that define a meaning for enclosed content, i.e. POST and PUT.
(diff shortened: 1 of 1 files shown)


manicminer (approved):
Thanks for this @magodo, this LGTM 👍

PR #434, opened 2023-04-26
Bugfix: configure AuxTenantIds for AzureCliAuthorizer

manicminer:
Also:

- Improve auth package testing, phase out MSI stub service and mock the HTTP clients for all authorizers instead.
- Unit test the `NewAuthorizerFromCredentials()` function
- Export `ClientAssertionAuthorizer` and `ClientSecretAuthorizer`
- Fix up license comment headers (align with project license)

Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/21346


sdk/auth/client.go
```diff
@@ -17,6 +17,26 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
+var (
+	// Client is the HTTP client used for sending authentication requests and obtaining tokens
+	Client HTTPClient
+
+	// MetadataClient is the HTTP client used for obtaining tokens from the Instance Metadata Service
+	MetadataClient HTTPClient
```
tombuildsstuff, line 25: do these need to be public?

tombuildsstuff (approved):
one comment but this otherwise LGTM

sdk/auth/client.go
manicminer: They do if we want to mock them out I think? But I think this will also address other potential needs for customizing the client(s), especially if you can shim them.

tombuildsstuff: I guess if we need them outside of this package, yeah fair point 👍

PR #445, opened 2023-05-03
client/sdk: support for `[]byte` requests and responses

tombuildsstuff:
This PR adds support for requests and responses containing `[]byte`'s - and also supports for marshaling and unmarshaling bodies containing powershell scripts which are defined as a byte array.

Supersedes #298

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -206,12 +206,8 @@ func (r *Response) Unmarshal(model interface{}) error {
 	}
 
 	if strings.Contains(contentType, "application/octet-stream") || strings.Contains(contentType, "text/powershell") {
-		val := reflect.ValueOf(model)
-		if val.Kind() != reflect.Pointer {
-			return fmt.Errorf("internal-error: `model` must be a pointer but it wasn't")
-		}
-		if _, ok := model.(*[]byte); !ok {
-			return fmt.Errorf("internal-error: `model` must be *[]byte but got %+v", model)
+		if _, ok := model.(**[]byte); !ok {
+			return fmt.Errorf("internal-error: `model` must be **[]byte but got %+v", model)
 		}
 
 		// Read the response body and close it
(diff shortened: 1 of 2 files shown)


jackofallops (approved):
Thanks @tombuildsstuff - LGTM 👍

sdk/client/client.go
```diff
@@ -217,13 +221,14 @@ func (r *Response) Unmarshal(model interface{}) error {
 		respBody = bytes.TrimPrefix(respBody, []byte("\xef\xbb\xbf"))
 
 		// copy the byte stream across
-		model = respBody
+		reflect.ValueOf(model).Elem().Elem().SetBytes(respBody)
```
wuxu92, line 224: we should not use  reflect to set the value of `**[]byte` type here, for it will cause panic when `model` is a zero valued pointer. why not use `*model = &respBody` directly?

```go
func SetBytes(model interface{}) error {
	if _, ok := model.(**[]byte); !ok {
		return fmt.Errorf("internal-error: `model` must be **[]byte but got %+v", model)
	}
	respBody := []byte("foo")
	reflect.ValueOf(model).Elem().Elem().SetBytes(respBody)
	return nil
}

func TestSetPtr(t *testing.T) {
	var p *[]byte
	if err := SetBytes(&p); err != nil {  <-- panic: reflect: call of reflect.Value.SetBytes on zero Value
		t.Fatal(err)
	}
	t.Logf("p value %v", *p)
}


////// ===========

func SetBytesV2(model interface{}) error {
	if pptr, ok := model.(**[]byte); !ok {
		return fmt.Errorf("internal-error: `model` must be **[]byte but got %+v", model)
	} else {
		respBody := []byte("foo")
		*pptr = &respBody
	}
	return nil
}

func TestSetPtrV2(t *testing.T) {
	var p *[]byte

	if err := SetBytesV2(&p); err != nil {
		t.Fatal(err)
	}
	t.Logf("p value %v", *p)  <-- set p as "foo" success
}
````

PR #448, opened 2023-05-04
sdk/client: the `Content-Length` header should be set for DELETE/PATCH requests

tombuildsstuff:
and match the size of the payload

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -320,20 +320,21 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 
 	var err error
 
-	// Ensure the Content-Lenght header is set for methods that define a meaning for enclosed content, i.e. POST and PUT.
-	// https://www.rfc-editor.org/rfc/rfc9110#section-8.6-5
-	if req.Method == "POST" || req.Method == "PUT" {
-		req.Header.Set("Content-Length", fmt.Sprintf("%d", req.ContentLength))
-	}
-
 	// Check we can read the request body and set a default empty body
 	var reqBody []byte
+	contentLength := 0
 	if req.Body != nil {
 		reqBody, err = io.ReadAll(req.Body)
 		if err != nil {
 			return nil, fmt.Errorf("reading request body: %v", err)
 		}
 		req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
+		contentLength = len(reqBody)
+	}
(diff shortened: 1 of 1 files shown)


jackofallops (approved):
LGTM 👍

sdk/client/client.go
```diff
@@ -320,20 +320,21 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
...
 		if err != nil {
 			return nil, fmt.Errorf("reading request body: %v", err)
 		}
 		req.Body = io.NopCloser(bytes.NewBuffer(reqBody))
+		contentLength = len(reqBody)
+	}
+	// Ensure the Content-Length header is set for methods that define a meaning for enclosed content, i.e. POST and PUT.
+	// https://www.rfc-editor.org/rfc/rfc9110#section-8.6-5
+	if strings.EqualFold(req.Method, http.MethodDelete) || strings.EqualFold(req.Method, http.MethodPatch) || strings.EqualFold(req.Method, http.MethodPost) || strings.EqualFold(req.Method, http.MethodPut) {
+		req.Header.Set("Content-Length", fmt.Sprintf("%d", contentLength))
```
tombuildsstuff, line 337: looks like this header is being stripped out when it's explicitly set to a `0` value, setting this (and the original logic) fails when it's set to `0`, but this works fine if another header is used

manicminer:
It seems like the stripping of the `Content-Length` and `Transfer-Encoding` headers is due to https://github.com/hashicorp/go-retryablehttp/pull/194 - neither of these headers need to be (or should be) manually set in the `Header` field, they are determined by `net/http` as part of the transport handling for HTTP/1.1 and HTTP/2.0. Accordingly, I think we should stop trying to set them manually (since it has no effect anyway due to them being automatically replaced by `net/http`) and wait for https://github.com/hashicorp/go-retryablehttp/pull/194.

manicminer (changes requested):
It seems like the stripping of the `Content-Length` and `Transfer-Encoding` headers is due to https://github.com/hashicorp/go-retryablehttp/pull/194 - neither of these headers need to be (or should be) manually set in the `Header` field, they are determined by `net/http` as part of the transport handling for HTTP/1.1 and HTTP/2.0. Accordingly, I think we should stop trying to set them manually (since it has no effect anyway due to them being automatically replaced by `net/http`) and wait for https://github.com/hashicorp/go-retryablehttp/pull/194.

PR #550, opened 2023-07-18
Environments: add Available() method for APIs

manicminer:
Related to https://github.com/hashicorp/terraform-provider-azurerm/pull/22400

### DIFF
--- a/sdk/environments/environment.go
+++ b/sdk/environments/environment.go
@@ -165,6 +165,10 @@ func (e *ApiEndpoint) withResourceIdentifier(identifier string) *ApiEndpoint {
 	return e
 }
 
+func (e *ApiEndpoint) Available() bool {
+	return e != nil && (e.resourceIdentifier != nil || e.endpoint != nil)
+}
+
 func (e *ApiEndpoint) DomainSuffix() (*string, bool) {
 	if e == nil {
 		return nil, false
@@ -187,6 +191,9 @@ func (e *ApiEndpoint) AppId() (*string, bool) {
 }
 
 func (e *ApiEndpoint) Name() string {
+	if e == nil {
+		return ""
+	}
 	return e.name
 }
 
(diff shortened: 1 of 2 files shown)


sdk/environments/environment.go
```diff
@@ -187,6 +191,9 @@ func (e *ApiEndpoint) AppId() (*string, bool) {
 }
 
 func (e *ApiEndpoint) Name() string {
+	if e == nil {
+		return ""
```
tombuildsstuff, line 195: potentially pedantic, but would help us diagnose when this is nil (initializing bug) vs left empty (logic bug)?

```suggestion
		return "(nil)"
```

sdk/environments/environment.go
```diff
@@ -165,6 +165,10 @@ func (e *ApiEndpoint) withResourceIdentifier(identifier string) *ApiEndpoint {
 	return e
 }
 
+func (e *ApiEndpoint) Available() bool {
+	return e != nil && (e.resourceIdentifier != nil || e.endpoint != nil)
```
tombuildsstuff, line 169: We'd need the Endpoint to know where to connect to, and the Resource Identifier to connect to it, so presumably should be:

```suggestion
	return e != nil && (e.resourceIdentifier != nil && e.endpoint != nil)
```

?

tombuildsstuff (changes requested):
Left a couple of comments inline but otherwise 👍

sdk/environments/environment.go
manicminer: Actually we can ignore the `resourceIdentifier` since we auto-fallback to use the `endpoint` e.g. in [scopes.go](https://github.com/hashicorp/go-azure-sdk/blob/b3ce3103af8e0d71698e7a16553bc7e06c796f9d/sdk/environments/scopes.go#L27-L40)

PR #553, opened 2023-07-19
sdk: skip unmarshalling for empty content type and 0 content length

stephybun:


sdk/client/client.go
```diff
@@ -151,6 +151,11 @@ func (r *Response) Unmarshal(model interface{}) error {
 
 	contentType := strings.ToLower(r.Header.Get("Content-Type"))
 	if contentType == "" {
+		// the maintenance API returns a 200 for a delete with no content-type and no content length so we should skip
+		// trying to unmarshal this
+		if r.ContentLength == 0 {
+			return nil
+		}
```
manicminer, line 158: I think this will be useful to have in all circumstances, since both `json.Unmarshal()` and `xml.Unmarshal()` return an EOFish error when receiving an empty byteslice. Let's move this to line 166? (i.e. before `if strings.Contains(contentType, "application/json") {`)

manicminer (approved):
Thanks for this fix @stephybun, this LGTM. Just one suggestion below, if we move this a bit further down it will help with more edge cases.

PR #568, opened 2023-07-27
sdk/client/dataplane/storage: updating the storage base client to take the baseUri rather than computing it

tombuildsstuff:
Storage has at least 3 known domain formats depending on the configuration of the Storage Account:

1. `https://{accountname}.{component}.blob.windows.net` - for generic storage accounts
2. `https://{accountname}.{dnszone}.{component}.storage.azure.net` - for storage accounts configured using AzureDNSZone
3. `https://{accountname}.{component}.{edgezone}.edgestorage.azure.net` - for storage accounts configured using an Edge Zone

These will also differ per environment too.

As such by updating the constructor to take the baseUri directly, we avoid issues with this.


manicminer (approved):
Makes sense, LGTM! 👍

PR #575, opened 2023-07-31
LRO poller: Adding more un-documented pending `status` for storage sync

### LINKED ISSUE #565: Issue with [StorageSync] / API Version [2020-03-01]
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Community Note

* Please vote on this issue by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Service Used

StorageSync

### API Versions Used

2020-03-01

### Description

Since [migrating `storage_sync` resources to `hashicorp/go-azure-sdk`](https://github.com/hashicorp/terraform-provider-azurerm/pull/21928), the acctests for storage sync and the alikes keep failing, e.g.:

```
------- Stdout: -------
=== RUN   TestAccStorageSync_update
=== PAUSE TestAccStorageSync_update
=== CONT  TestAccStorageSync_update
    testcase.go:113: Step 5/6 error: Error running apply: exit status 1
        Error: creating Storage Sync Service (Subscription: "*******"
        Resource Group Name: "acctestRG-ss-230725005007008353"
        Storage Sync Service Name: "acctest-SS-230725005007008353"): polling after StorageSyncServicesCreate: `result.Status` was nil/empty - `op.Status` was "finishNewStorageSyncService" / `op.Properties.ProvisioningState` was ""
          with azurerm_storage_sync.test,
          on terraform_plugin_test.tf line 24, in resource "azurerm_storage_sync" "test":
          24: resource "azurerm_storage_sync" "test" {
--- FAIL: TestAccStorageSync_update (403.32s)
FAIL
```

The reason is as seen in the above error message, that the service API returns some unknown `status`, which doesn't belong to any of the status expected by `longRunningOperationPoller`:

https://github.com/magodo/terraform-provider-azurerm/blob/main/vendor/github.com/hashicorp/go-azure-sdk/sdk/client/resourcemanager/poller_lro.go#L152-L171.

In contrast, the Track1 SDK has below logic for polling:

https://github.com/magodo/terraform-provider-azurerm/blob/d6bc225707d78a51f7db65824bb84aab197aa536/vendor/github.com/Azure/go-autorest/autorest/azure/async.go#L570-L575

```go
		if pt.resp.StatusCode == http.StatusAccepted {
			pt.State = operationInProgress
		} else if provStateApl {
			if ps := pt.getProvisioningState(); ps != nil {
				pt.State = *ps
			} else {
				pt.State = operationSucceeded
			}
		} else {
			return autorest.NewError("pollingTrackerBase", "updatePollingState", "the response from the async operation has an invalid status code")
		}
```

In this case, any unknown status will be regarded as succeeded as long as the response status code is successful. This somehow hides the issue, whilst I don't think it is the right way.

So my question is do I need to extend the "known" status for the `longRunningOperationPoller` to include the reported status?

The problem is that, from the test failures, so far we have following status for storage sync resource:

- finishNewStorageSyncService
- validateInput
- newPrivateDnsEntries

And we have following status for the storage sync cloud endpoint resource:

- newReplicaGroup

The swagger doesn't define the full list of status, where the `status` is defined as a plain string.

### References

_No response_

magodo:
Fix #565

### DIFF
--- a/sdk/client/resourcemanager/poller_lro.go
+++ b/sdk/client/resourcemanager/poller_lro.go
@@ -167,6 +167,14 @@ func (p *longRunningOperationPoller) Poll(ctx context.Context) (result *pollers.
 			"Running": pollers.PollingStatusInProgress,
 			// KubernetesConfiguration@2022-11-01 returns `Updating` rather than `InProgress` during update
 			"Updating": pollers.PollingStatusInProgress,
+			// StorageSync@2020-03-01 returns `validateInput`, `newPrivateDnsEntries`, `finishNewStorageSyncService` rather than `InProgress` during creation/update
+			// See: https://github.com/hashicorp/go-azure-sdk/issues/565
+			"validateInput":               pollers.PollingStatusInProgress,
+			"newPrivateDnsEntries":        pollers.PollingStatusInProgress,
+			"finishNewStorageSyncService": pollers.PollingStatusInProgress,
+			// StorageSync@2020-03-01 (CloudEndpoints) returns `newReplicaGroup` rather than `InProgress` during creation/update
+			// See: https://github.com/hashicorp/go-azure-sdk/issues/565
+			"newReplicaGroup": pollers.PollingStatusInProgress,
 		}
 		for k, v := range statuses {
 			if strings.EqualFold(string(op.Properties.ProvisioningState), string(k)) {


manicminer (approved):
Thanks @magodo, we might want to look at making these into custom pollers in the long run but I think this is OK for now 👍

PR #583, opened 2023-08-01
sdk/client/resourcemanager: support for LROs which also return immediately

tombuildsstuff:
Both API Management/Schema @ 2021-08-01/2022-08-01 and Automation/DSCConfiguration@2020-01-13-preview support returning both a LRO and an immediate response. In this case since we retrieve the original URI and get a 200 OK /without/ a provisioningState header - we'll special-case this for now.

Fixes https://github.com/hashicorp/go-azure-sdk/issues/542


manicminer (approved):
Thanks @tombuildsstuff, LGTM 🏟️

Issue #492, opened 2023-05-29
`Migrate` @ `2020-01-01` - support for nonstandard next link


ziyeqf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Community Note

* Please vote on this issue by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Service Used

Migrate

### API Versions Used

2020-01-01

### Description

The response is paged with [`nextLink`](https://github.com/Azure/azure-rest-api-specs/blob/a13f142cfd0c2273b9a9d27dc82babcd91d7d1a1/specification/migrate/resource-manager/Microsoft.OffAzure/stable/2020-01-01/migrate.json#L3685) not in OData and sdk ignored it.

### References

Swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/migrate/resource-manager/Microsoft.OffAzure/stable/2020-01-01/migrate.json#L3685

tombuildsstuff:
@ziyeqf that's a Swagger bug which would need to be fixed - would you mind sending a PR to fix that?

ziyeqf:
thanks @tombuildsstuff, but I may not get the point. If I submit a PR to rename `nextLink` to `@odata.nextlink`, the service will still return in the original format?

tombuildsstuff:
@ziyeqf the Swagger should be defining `x-ms-pageable` to state which field contains the `nextLink`, see here: https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-pageable

As such just updating the API definition should be sufficient there, I don't think we need a Pandora change for this, I'm kinda surprised there isn't a linter to catch a field containing `nextLink` without a `x-ms-pageable` defined actually?

tombuildsstuff:
Got it, thanks - in which case yeah this is a bug in the new base layer here where we're relying on the OData value rather than using the value we're expecting:

https://github.com/hashicorp/go-azure-sdk/blob/af92233a172a86b8bddc32d55ed22979979f29df/sdk/client/client.go#LL466C1-L483C1

@manicminer mind taking a look into that one?

ziyeqf:
[...]

PR #587, opened 2023-08-01
Base Client: support custom paging

manicminer:
Fixes: #492

Importer/generator will need to be updated to output custom pagers for APIs that require it.


tombuildsstuff (approved):
A couple of comments but otherwise 👍

PR #590, opened 2023-08-02
client/resourcemanager: handling self-referenced LROs

tombuildsstuff:
Some LROs return a 201 with the Location header as a self-reference, polling on that URI returns a 200 OK with no provisioningState or status field to poll on - as such we need to special-case to handle this.

Fixes an issue in `Automation` @ `2022-08-08` in `DSCNodeConfiguration` / `CreateOrUpdate`

Fixes https://github.com/hashicorp/terraform-provider-azurerm/pull/22781


manicminer (approved):
LGTM 🧇

PR #597, opened 2023-08-03
sdk/client/resourcemanager: supporting 202's with a self reference

tombuildsstuff:
Most of API Management returns a 202 with a self reference containing an `asynctoken`, as such if a self-reference exists then we can use this URI directly.

Fixes:

```
Error: creating/updating Api (Subscription: "*******"
        Resource Group Name: "acctestRG-230803055317928116"
        Service Name: "acctestAM-230803055317928116"
        Api: "acctestAMA-230803055317928116"): performing CreateOrUpdate: no applicable pollers were found for the response
```

From https://github.com/hashicorp/terraform-provider-azurerm/pull/22783


manicminer (approved):
Nice, LGTM 👍

PR #611, opened 2023-08-10
`sdk/client`: req marshal with `[]byte` instead of `*[]byte`

wuxu92:
The argument for request marshal is `[]byte` not '`*[]byte` (the response is `*[]byte`)

so method below would failed with err: `performing ReplaceContent: internal-error: `payload` must be *[]byte but got [xx xx]`

https://github.com/hashicorp/go-azure-sdk/blob/32fa38105db88915b1a53cc2a590056821989a88/resource-manager/automation/2022-08-08/runbookdraft/method_replacecontent.go#L24-L40

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -109,13 +109,16 @@ func (r *Request) Marshal(payload interface{}) error {
 	}
 
 	if strings.Contains(contentType, "application/octet-stream") || strings.Contains(contentType, "text/powershell") {
-		v, ok := payload.(*[]byte)
-		if !ok {
+		switch v := payload.(type) {
+		case *[]byte:
+			r.ContentLength = int64(len(*v))
+			r.Body = io.NopCloser(bytes.NewReader(*v))
+		case []byte:
+			r.ContentLength = int64(len(v))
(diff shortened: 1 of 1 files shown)


sdk/client/client.go
```diff
@@ -109,13 +109,16 @@ func (r *Request) Marshal(payload interface{}) error {
...
-			return fmt.Errorf("internal-error: `payload` must be *[]byte but got %+v", payload)
+		switch v := payload.(type) {
+		case *[]byte:
+			r.ContentLength = int64(len(*v))
+			r.Body = io.NopCloser(bytes.NewReader(*v))
+		case []byte:
+			r.ContentLength = int64(len(v))
+			r.Body = io.NopCloser(bytes.NewReader(v))
+		default:
+			return fmt.Errorf("internal-error: `payload` must be *[]byte but got type %T", payload)
```
manicminer, line 120:
```suggestion
			return fmt.Errorf("internal-error: `payload` must be []byte or *[]byte but got type %T", payload)
```

manicminer (approved):
Thanks for this @wuxu92, this looks ok to me, I would just make sure to update the error message for clarity.

PR #618, opened 2023-08-14
environments: add `IoTCentral` subdomain to public environment

stephybun:


sdk/environments/helpers.go
```diff
@@ -85,6 +85,16 @@ func DataLakeAPI(domainSuffix string) Api {
 	}
 }
 
+func IoTCentral(domainSuffix string) Api {
+	return &ApiEndpoint{
+		domainSuffix:       pointer.To(domainSuffix),
+		endpoint:           nil,
+		appId:              pointer.To(iotCentralAppId),
+		name:               "IoTCentral",
+		resourceIdentifier: nil,
```
tombuildsstuff, line 94: this is used to obtain an authorization token, so would be needed (and will be the same in all environments, fwiw)

tombuildsstuff (approved):

PR #623, opened 2023-08-16
`sdk/client` : allow content-type to not be set

catriona-m:
Changing this after coming across this request which doesn't use a content-type: https://learn.microsoft.com/en-us/rest/api/storageservices/get-file


tombuildsstuff (approved):

tombuildsstuff:
@manicminer think it's worth a Resource Manager/Graph specific check to check for this one?

manicminer:
Makes sense, I think we should consider moving this check to the Resource Manager & Graph clients rather than remove it altogether, WDYT @catriona-m?

tombuildsstuff (approved):
LGTM 👍

PR #638, opened 2023-08-30
`sdk/client`: add xml header to marshalled payload

catriona-m:


sdk/client/client.go
```diff
@@ -102,8 +102,9 @@ func (r *Request) Marshal(payload interface{}) error {
 	if strings.Contains(contentType, "application/xml") || strings.Contains(contentType, "text/xml") {
 		body, err := xml.Marshal(payload)
 		if err == nil {
-			r.ContentLength = int64(len(body))
-			r.Body = io.NopCloser(bytes.NewReader(body))
+			bodyWithHeader := []byte(xml.Header + string(body))
```
tombuildsstuff, line 105: Apologies for the delayed review - thinking about this one the SDK shouldn't be mutating the payload directly, so the caller should be adding the XML header if required - as such could we do this within Giovanni, rather than within the SDK?

manicminer: Agreed, if necessary this could also be done in the intermediate client rather than the base client?

tombuildsstuff: Good shout to wrap the `Marshal` function in [the Storage Base Client](https://github.com/hashicorp/go-azure-sdk/blob/main/sdk/client/dataplane/storage/client.go) - FWIW there _are_ a handful of XML and YAML instances within ARM (APIM?), but in both of those cases I think we'd handle that within the Provider instead

catriona-m:
I made these changes directly in giovanni instead, so going to close this now, thanks for reviewing!

PR #665, opened 2023-09-26
auth: manual token expiration, better auxiliary token handling, improve tests

manicminer:
## Better Handling of Auxiliary Tokens

Adds the `AuthorizeRequest` field to `client.Client` which API clients can define for customized authoriation behavior. This is then utilized by `resourcemanager.Client` to decorate requests with both `Authorization` and `x-ms-authorization-auxiliary` headers.

Adds a helper function `SetAuthHeader()` which decorates an `*http.Request` with a bearer token, e.g. adds the `Authorization` header. This is called by `base.Client` when `AuthorizeRequest` is nil.

## Manual Token Expiration

This provides the ability to intentionally expire a cached access token, which causes a new token to be automatically acquired the next time the `Token()` or `AuxiliaryTokens()` methods are called.

Downstream implementers can test whether an `auth.Authorizer` implements `auth.CachingAuthorizer`, in case they receive a standard `auth.Authorizer`, and then call `InvalidateTokenCache()` prior to making a new request. This requires that the method is called as-needed when an action is taken that is likely to necessitate a new access token, for example with management groups which make use of token claims for access control.

Closes: #459 

Related:
- https://github.com/hashicorp/terraform-provider-azurerm/issues/12478
- https://github.com/hashicorp/terraform-provider-azurerm/issues/15894
- https://github.com/hashicorp/terraform-provider-azurerm/issues/19595
- https://github.com/hashicorp/terraform-provider-azurerm/issues/23238

### DIFF
--- a/sdk/client/resourcemanager/authorization.go
+++ b/sdk/client/resourcemanager/authorization.go
@@ -37,7 +37,9 @@ func AuthorizeResourceManagerRequest(ctx context.Context, req *http.Request, aut
 	for _, auxToken := range auxTokens {
 		auxTokenValues = append(auxTokenValues, fmt.Sprintf("%s %s", auxToken.Type(), auxToken.AccessToken))
 	}
-	req.Header.Set("X-Ms-Authorization-Auxiliary", strings.Join(auxTokenValues, ", "))
+	if len(auxTokenValues) > 0 {
+		req.Header.Set("X-Ms-Authorization-Auxiliary", strings.Join(auxTokenValues, ", "))
+	}
 
 	return nil
 }


manicminer:
Unit test results

Screenshot 2023-09-27 at 13 04 58

industrialzombie:
Hey,

The `client` will need to be updated to use this new `SetAuthHeaders()` function when it executes a request.
Should that be included in the scope of this change?

https://github.com/hashicorp/go-azure-sdk/blob/main/sdk/client/client.go#L337

manicminer:
@industrialzombie Yes it will, I had thought to add that separately for additional safety downstream, but on balance it's probably fine to add it here and merge it all together 👍

industrialzombie:
Is it possible to get an estimate on when this might be reviewed?

manicminer:
@industrialzombie We'll look at this after today's release, so that we have a few days to catch any potential bugs before shipping.

industrialzombie:
Thanks, I appreciate the feedback!

sdk/auth/cached_authorizer.go
```diff
@@ -67,6 +68,15 @@ func (c *CachedAuthorizer) AuxiliaryTokens(ctx context.Context, req *http.Reques
 	return c.auxTokens, nil
 }
 
+// ExpireTokens expires the currently cached token and auxTokens, forcing new tokens to be acquired when Token() or AuxiliaryTokens() are next called
+func (c *CachedAuthorizer) ExpireTokens() error {
+	c.token.Expiry = time.Now()
```
tombuildsstuff, line 73: sanity-checking: what if `c.Token` is nil here?

sdk/auth/interface.go
```diff
@@ -13,8 +13,8 @@ import (
 // Authorizer is anything that can return an access token for authorizing API connections
 type Authorizer interface {
 	Token(ctx context.Context, request *http.Request) (*oauth2.Token, error)
-
 	AuxiliaryTokens(ctx context.Context, request *http.Request) ([]*oauth2.Token, error)
+	ExpireTokens() error
```
tombuildsstuff, line 17: for this one (but also in retrospect) is it worth adding some documentation for this one? As it stands the "will do nothing" comments on each implementation will show as needed, but don't strictly explain the purpose for this one?

sdk/auth/token.go
```diff
@@ -4,15 +4,54 @@
...
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-azure-sdk/sdk/claims"
+	"golang.org/x/oauth2"
 )
 
+// SetAuthHeaders decorates a *http.Request with necessary authorization headers for Azure APIs. For more information about the vendor-specific
+// `x-ms-authorization-auxiliary` header, see https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant
+func SetAuthHeaders(ctx context.Context, req *http.Request, authorizer Authorizer) error {
```
tombuildsstuff, line 19: since this is specific to Resource Manager, this probably wants to go into the Resource Manager Client?

sdk/auth/token.go
```diff
@@ -26,7 +65,11 @@ func tokenDueForRenewal(token *oauth2.Token) bool {
 	expiry := token.Expiry.Round(0)
 	delta := tokenExpiryDelta
 	now := time.Now()
-	expiresWithinTenMinutes := expiry.Add(-delta).Before(now)
+
+	// Always return early if the token validity doesn't extend past the expiry delta
+	if expiry.Add(-delta).Before(now) {
```
tombuildsstuff, line 70: There's a `sub`tract we can use here:

```suggestion
	if expiry.Sub(delta).Before(now) {
```

sdk/client/client.go
```diff
@@ -327,14 +327,11 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 		return nil, fmt.Errorf("req.Request was nil")
 	}
 
-	// at this point we're ready to send the HTTP Request, as such let's get the Authorization token
-	// and add that to the request
+	// Set Authorization and X-Ms-Authorization-Auxiliary headers
```
tombuildsstuff, line 330: Doing this will cause issues with Storage Data Plane which is _super_ picky about the headers it supports

Perhaps Client should have a `ConfigureAuthorizerForRequest` function/method defined against the struct which the implementations can pass-in?

tombuildsstuff, line 330: Doing this will cause issues with Storage Data Plane which is _super_ picky about the headers it supports

Perhaps Client should have a `ConfigureAuthorizerForRequest` function/method defined against the struct which the implementations can pass-in?

sdk/auth/shared_key_authorizer.go
```diff
@@ -64,7 +64,10 @@ func (s *SharedKeyAuthorizer) AuxiliaryTokens(_ context.Context, _ *http.Request
 	return []*oauth2.Token{}, nil
 }
 
-// ---
+// ExpireToken has no effect with shared keys
+func (c *SharedKeyAuthorizer) ExpireTokens() error {
+	return fmt.Errorf("SharedKeyAuthorizer tokens cannot expire")
```
tombuildsstuff, line 69: given the other authorizers are a noop, why not do the same here? perhaps this method wants renaming to `InvalidateTokenCache` or something to better reflect what it's doing?

tombuildsstuff (approved):
Left some comments inline but otherwise 👍

sdk/auth/cached_authorizer.go
manicminer: thanks, good catch

sdk/auth/interface.go
manicminer: good shout, will add 👍

sdk/auth/token.go
manicminer: agreed, other APIs may support the concept of multiple tenants but the implementation will be different

sdk/auth/shared_key_authorizer.go
manicminer: I've done away with this ambiguity and added a `CachingAuthorizer` interface which can be tested for at runtime to call the `InvalidateTokenCache()` method.

sdk/client/client.go
manicminer: Makes sense, I've added this and set a custom authorization function for `ResourceManagerClient`. Where such custom function is not specified, the base client will fall back to setting a standard `Authorization` header with a bearer token (i.e. its current behavior).

sdk/auth/token.go
manicminer: `Sub` actually does the inverse and subtracts two `Time` structs to return a `Duration`

manicminer:
Thanks for the review @tombuildsstuff, I've reworked it to remove the anbiguity over invalidation by adding a `CachingAuthorizer` interface, and added an `AuthorizeRequest` property to the base client to allow API clients to supply a custom function for decorating the request. WDYT?

tombuildsstuff (approved):
A couple of minor comments but otherwise 👍

PR #673, opened 2023-10-11
tooling: support for conditionally updating the AzureRM Provider via a Tag

tombuildsstuff:
This PR (attempts) to conditionally update the dependency in `hashicorp/terraform-provider-azurerm` whenever a release occurs in this repository, and the release pull request includes `update-azurerm-after-release`. As mentioned in https://github.com/hashicorp/terraform-provider-azurerm/pull/23498, the tool being run determines the changes in the latest SDK release and then attempts to upgrade each of the AzureRM Services to using this version.

This PR also updates the GHA's to pull the Go version from the `.go-version` file - since they weren't before.

There's an intentional sleep in this script since we need to wait for the GHA on the `hashicorp/terraform-provider-azurerm` repository to be run (to open the PR) - ~which also needs to be added before this PR can be merged~ done (~as do permissions~ done).

Dependent on https://github.com/hashicorp/terraform-provider-azurerm/pull/23498

### DIFF
--- a/.github/workflows/automation-release.yaml
+++ b/.github/workflows/automation-release.yaml
@@ -13,6 +13,9 @@ jobs:
     runs-on: custom-linux-medium
     permissions:
       contents: write
+    outputs:
+      latest_tag: ${{ steps.results.outputs.latest_tag }}
+      should_update_azurerm: ${{ steps.results.outputs.should_update_azurerm }}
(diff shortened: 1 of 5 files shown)


manicminer (approved):
Awesome, LGTM! 👍

PR #685, opened 2023-10-18
automation: the output value is a literal boolean

tombuildsstuff:
Right now this is a string, which means these aren't matching


tombuildsstuff:
Again skipping review since this is a nit

PR #691, opened 2023-10-20
github actions: being explicit with the tag

tombuildsstuff:
This isn't being expanded, so doing this as a workaround


tombuildsstuff:
This is a minor/CI fix - skipping review

PR #693, opened 2023-10-23
automation: updating the release script to use bash explicitly

tombuildsstuff:


tombuildsstuff:
This is a (minor) bug fix for GHA, skipping review since this can only be tested in GHA and is pass/fail

PR #694, opened 2023-10-23
automation: fixing the step name

tombuildsstuff:


tombuildsstuff:
Skipping review since this is a minor nit to fix GHA (that can only be tested here)

PR #696, opened 2023-10-23
automation: checking out the full history for the release step

tombuildsstuff:
This is required in order to obtain the diff between the previous tag and current tag during the release/when updating the AzureRM Provider


tombuildsstuff:
Minor GHA bugfix, skipping review since this can only be tested in GHA

PR #697, opened 2023-10-23
automation: updating the release script to checkout for the patcher step

tombuildsstuff:
The checkouts aren't reused, plus also fixing an issue where the description wouldn't be exported correctly


tombuildsstuff:
Skipping review since this is another nit that has to go through a PR to get triggered correctly/can only be tested in GHA - it's a shame these can't be tested locally.

PR #699, opened 2023-10-23
automation: the email has to be escaped using single and not double quotes

tombuildsstuff:
This differs from the usage in Pandora, but solves the issue, so fine.


tombuildsstuff:
Super minor nit, tested in an isolated environment - skipping review to confirm in GHA

PR #707, opened 2023-10-26
Avoid repeating "az version" executions

orgads:
There is no reason to execute it more than once, and at least on Windows, each process has a noticable cost.

### DIFF
--- a/sdk/internal/azurecli/azcli.go
+++ b/sdk/internal/azurecli/azcli.go
@@ -14,17 +14,20 @@ import (
 	"github.com/hashicorp/go-version"
 )
 
+var cliVersion *struct {
+	AzureCli          *string      `json:"azure-cli,omitempty"`
+	AzureCliCore      *string      `json:"azure-cli-core,omitempty"`
+	AzureCliTelemetry *string      `json:"azure-cli-telemetry,omitempty"`
+	Extensions        *interface{} `json:"extensions,omitempty"`
+}
+
 // CheckAzVersion tries to determine the version of Azure CLI in the path and checks for a compatible version
 func CheckAzVersion(minVersion string, nextMajorVersion *string) error {
-	var cliVersion *struct {
-		AzureCli          *string      `json:"azure-cli,omitempty"`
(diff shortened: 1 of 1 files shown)


manicminer:
Hi @orgads, thanks for suggesting this change. We've looked into this to double check, and whilst it seems that this would cache the result of the `az version` command, it won't reduce the number of times this command is run.

We already provide the ability to cache access tokens after acquiring them (even when acquiring them from Azure CLI), with the `CachedAuthorizer` struct. This is designed to minimize the number of times that a backend (e.g. `az` command) is called and means that the Azure CLI is only invoked when a new token is needed. Additionally, the `az version` command is run once per token acquisition per API scope, so in effect we're already calling it the minimum number of times. We do want to call `az version` each time in order to catch the scenario where the Azure CLI version changes during the validity period of an access token (which is typically just under one hour).

If you are running Terraform, I believe the reason you may be seeing this many times is due to the way that Terraform handles parallelism. A provider may be executed multiple times in parallel, and each instance of a provider is unaware of the other instance(s) and does not share memory. Therefore, the token caching that takes place is limited to that provider instance, but will take effect for the lifetime of that access token. I believe that access tokens for additional data plane API scopes are requested as-needed, so these should not compound the startup time of the provider.

Accordingly, I'm going to close this PR for now, as the proposal doesn't offer additional capability beyond the `CachedAuthorizer` struct already present in the package. Thank you again for looking into this and for opening the PR.

orgads:
Thanks for the detailed explanation.

orgads:
@manicminer I only see one instance of the provider, which keeps running `az version` over and over again. Maybe something is wrong with the cache? Is there a way to trace that?

manicminer:
@orgads With further testing, I can corroborate what you're seeing and on balance I believe further thought is needed on improving performance here. It's also not just `az version` - we may also need to look at `az account show` which is called repeatedly by the TF providers. Further complicating this that by default, the Azure CLI emits telemetry to Application Insights whenever it's run, further slowing things down.

orgads:
Better replace this PR with an issue then, right?

orgads:
Done:
- #752

PR #715, opened 2023-11-06
Changing the cloud names to be the same as the one defined in the metadata host

magodo:
This PR renames the Public/USGov/China cloud environment names to the ones defined in their respective metadata host (2022-09-01 schema).

The motivation of the renaming is that in the azurerm provider, there are if checks for the current environment name, in which case, the name will be different when directly specified, i.e. via the `environment` property (e.g. it is `public` for public cloud), or read from the metadata host (e.g. it is `AzureCloud` for public cloud).

Example code: https://github.com/hashicorp/terraform-provider-azurerm/blob/41be3be3b9621c0f40014a172a678227317950f8/internal/services/storage/storage_account_resource.go#L1150 (related issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/23764)

Meanwhile, it enables the `TestFromEndpoint_USGovernment` test case given the 2022-09-01 API version works now.

## Test

```shell
💢  go test -v ./...
=== RUN   TestIsAzureStack_BuiltInEnvironmentsShouldNotBeMarkedAsStack
2023/11/06 11:40:03 Environment: "AzureCloud"
2023/11/06 11:40:03 Environment: "Canary"
2023/11/06 11:40:03 Environment: "AzureChinaCloud"
2023/11/06 11:40:03 Environment: "AzureUSGovernment"
2023/11/06 11:40:03 Environment: "USGovernmentL5"
--- PASS: TestIsAzureStack_BuiltInEnvironmentsShouldNotBeMarkedAsStack (0.00s)
=== RUN   TestIsAzureStack
2023/11/06 11:40:03 Environment "AzureStackCloud"
2023/11/06 11:40:03 Environment "Identity Provider Empty"
2023/11/06 11:40:03 Environment "Identity Provider Custom"
2023/11/06 11:40:03 Environment "Tenant Empty"
2023/11/06 11:40:03 Environment "Tenant Custom"
--- PASS: TestIsAzureStack (0.00s)
=== RUN   TestEndpointRefresh_China
--- PASS: TestEndpointRefresh_China (0.37s)
=== RUN   TestEndpointRefresh_Public
--- PASS: TestEndpointRefresh_Public (0.57s)
=== RUN   TestEndpointRefresh_USGovernment
--- PASS: TestEndpointRefresh_USGovernment (1.63s)
=== RUN   TestFromEndpoint_Public
--- PASS: TestFromEndpoint_Public (0.51s)
=== RUN   TestFromEndpoint_USGovernment
--- PASS: TestFromEndpoint_USGovernment (1.17s)
=== RUN   TestFromEndpoint_China
--- PASS: TestFromEndpoint_China (0.09s)
=== RUN   TestFromName
    from_name_test.go:54: Printing "canary"
    from_name_test.go:54: Printing "china"
    from_name_test.go:54: Printing "dod"
    from_name_test.go:54: Printing "global"
    from_name_test.go:54: Printing "public"
    from_name_test.go:54: Printing "usgovernment"
    from_name_test.go:54: Printing "usgovernmentl4"
    from_name_test.go:54: Printing "usgovernmentl5"
    from_name_test.go:54: Printing "mars"
--- PASS: TestFromName (0.00s)
PASS
ok      github.com/hashicorp/go-azure-sdk/sdk/environments      4.356s
```

### DIFF
--- a/sdk/environments/azure_china.go
+++ b/sdk/environments/azure_china.go
@@ -3,7 +3,7 @@
 
 package environments
 
-const AzureChinaCloud = "China"
+const AzureChinaCloud = "AzureChinaCloud"
 
 func AzureChina() *Environment {
 	env := baseEnvironmentWithName(AzureChinaCloud)
--- a/sdk/environments/azure_gov.go
+++ b/sdk/environments/azure_gov.go
@@ -3,7 +3,7 @@
 
 package environments
 
-const AzureUSGovernmentCloud = "USGovernment"
+const AzureUSGovernmentCloud = "AzureUSGovernment"
 
 func AzureUSGovernment() *Environment {
 	env := baseEnvironmentWithName(AzureUSGovernmentCloud)
--- a/sdk/environments/azure_public.go
+++ b/sdk/environments/azure_public.go
@@ -3,7 +3,7 @@
 
 package environments
 
-const AzurePublicCloud = "Public"
+const AzurePublicCloud = "AzureCloud"
 
 func AzurePublic() *Environment {
 	env := baseEnvironmentWithName(AzurePublicCloud)
--- a/sdk/environments/from_endpoint_test.go
+++ b/sdk/environments/from_endpoint_test.go
@@ -14,20 +14,29 @@ func TestFromEndpoint_Public(t *testing.T) {
 		t.Fatalf("loading from endpoint: %+v", err)
 	}
 
-	if actual.Name != "AzureCloud" {
-		t.Fatalf("expected the Environment name to be `AzureCloud` but got %q", actual.Name)
+	if actual.Name != AzurePublicCloud {
+		t.Fatalf("expected the Environment name to be `%s` but got %q", AzurePublicCloud, actual.Name)
 	}
 }
 
 func TestFromEndpoint_USGovernment(t *testing.T) {
-	t.Skip("Skipping because USGovernment ARM metadata service has reverted to older schema 2019-05-01 (manicminer, 2023-02-02)")
-
 	actual, err := FromEndpoint(context.Background(), "https://management.usgovcloudapi.net", "AzureUSGovernment")
 	if err != nil {
 		t.Fatalf("loading from endpoint: %+v", err)
 	}
 
-	if actual.Name != "AzureCloud" {
-		t.Fatalf("expected the Environment name to be `AzureCloud` but got %q", actual.Name)
+	if actual.Name != AzureUSGovernmentCloud {
+		t.Fatalf("expected the Environment name to be `%s` but got %q", AzureUSGovernmentCloud, actual.Name)
+	}
+}
+
+func TestFromEndpoint_China(t *testing.T) {
+	actual, err := FromEndpoint(context.Background(), "https://management.chinacloudapi.cn", "AzureUSGovernment")
+	if err != nil {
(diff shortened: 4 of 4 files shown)


tombuildsstuff:
@magodo FWIW [this bug stems from comparing the two different environment objects within the provider](https://github.com/hashicorp/terraform-provider-azurerm/blob/41be3be3b9621c0f40014a172a678227317950f8/internal/services/storage/storage_account_resource.go#L1150) so we'll want to fix this there instead.

We're intentionally using different names for the environments in `hashicorp/go-azure-sdk` from those defined in `Azure/go-autorest` to both catch issues like this - since in (almost) all cases where we're using the `Environment` object we should be using the `IsAvailable()` method rather than comparing on the name - but also to allow this to become a user-presentable value in the future.

In this case I think there's two separate issues here:

1. The Environments used for comparison in the Storage Account resource don't match
2. The `metadata_host` field isn't expected to be set to the default value of an Environment

Which I think want to be fixed in 3 ways:

1. The `Environment` struct wants a `.Matches(other Environment) bool` function, which can today compare by name - but would allow us to compare on additional fields in the future.
2. The updated `.Matches(other Environment) bool` function wants to be used in place of the comparison within the Storage Account (and any other places that use an `Environment` check that isn't calling `.IsAvailable()`).
3. The provider should raise an error when `metadata_host` is set to the default value for a default environment (e.g. `management.azure.com` for public, etc) - since this field should instead be omitted from the config.

Whilst that last one may seem excessive, that field is intended to be set only in a handful of niche circumstances (e.g. when using a custom Azure Environment, or when targeting a region specific endpoint - although the Provider may do that itself in the future) - so we should explicitly discourage this from being set for a known Azure Environment.

As such I don't think we should be updating the `name` value for these environments, but instead the changes outlined above should allow us to avoid this category of issues going forwards - WDYT?

magodo:
@tombuildsstuff I think the comparisom [here](https://github.com/hashicorp/terraform-provider-azurerm/blob/41be3be3b9621c0f40014a172a678227317950f8/internal/services/storage/storage_account_resource.go#L1150) is comparing the same thing (i.e. the `Environment.Name` from `github.com/hashicorp/go-azure-sdk/sdk/environments/Environment`).

Regarding the `IsAvailable()`, I didn't find it is defined for the `Environment` type. I'm also not sure whether the `Matches()` method is a good idea here, since it is not clear whether one environment matches another one, there is more of *is*/*isn't*, instead of *matches*?

For the additional restriction of the `metadata_host` in the provider config, it indeed seems weired to me to disallow users to specify its *default* value.

Overall, I don't see an outstanding issue to not compare environment by its name, and make the name to be the same as what is defined in the metadata host.

manicminer:
Just to follow up on this one. The linked usage in `azurerm_storage_account` has since been removed as it was an impossible comparison - the particular statement was checking to see if an unrecognized cloud was in use (e.g. any private cloud); since these are by definition unknown, it was pointless to try and assert available functionality there.

At the same time as removing the above statement, I audited the provider and found no other valid use cases where we need to determine or assert some available functionality for a given cloud. Additionally, such assertions would always be invalid for private clouds. Given that there is no use case at this time, I suggest we revisit this on an as-needed basis rather than try to design this upfront.

As such, I'll close this issue for now and we can revisit when the need is less hypothetical :)

PR #724, opened 2023-11-08
automation: base64encoding/decoding the pr description

tombuildsstuff:
This spans multiple lines, so base64-encoding this is a delightful workaround to allow us to pass this between stages

There is a size limit to the amount of data that can be passed, but it's unlikely to exceed 1MB so it's probably fine


tombuildsstuff:
This is another minor change to the automation, since this has to be triggered via a PR and is pass/fail, I'm gonna skip review for this one.

PR #725, opened 2023-11-08
automation: wrapping the description as a multi-line string

tombuildsstuff:


tombuildsstuff:
This should force the output to be treated as a multi-line string, so this can be encoded/decoded correctly - but again since this is pass/fail related to the automation - I'm going to skip review for this one

PR #728, opened 2023-11-09
automation: combining the release and comment steps into one

tombuildsstuff:
This isn't ideal, as it would be nice to have the ability to rerun the comment stage, but the credentials are scoped so this is fine.


tombuildsstuff:
As previously this is pass/fail, and has been discussed with @jackofallops so I'm skipping review here

PR #729, opened 2023-11-09
automation: updating the command used to identify the PR Number

tombuildsstuff:
This needs to escape the command to be able to run successfully


tombuildsstuff:
This is another pass/fail that can only be really tested in GHA - running this locally (without the substitution) works fine so skipping review.

PR #732, opened 2023-11-09
automation: retrying when getting the PR number from Github

tombuildsstuff:
This allows extra time if the other GHA takes a while


tombuildsstuff:
This is an automation change that can only be tested in GHA that's been discussed offline - skipping review

PR #733, opened 2023-11-09
automation: updating the polling interval

tombuildsstuff:
Polling once every 15s, for a maximum of 20 times - to allow for the PR to be opened in the background


tombuildsstuff:
This is a minor automation change that can only be tested in GHA - skipping review since this is pass/fail

Issue #752, opened 2023-11-28
az version (and others) is called many times when running terraform azurerm provider


orgads:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Community Note

* Please vote on this issue by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Service Used

Any

### API Versions Used

Any

### Description

When Terraform azurerm provider is used, `az version` and some other commands are executed many times during the execution.

### References

Initially discussed in:
- #707

orgads:
@manicminer I still see many invocations of azure cli for get-access-token, with the same arguments. Can this be cached too?

```
python  -IBm azure.cli account get-access-token --scope https://management.azure.com/.default --subscription <subscription-id> -o=json
```

PR #753, opened 2023-11-28
auth: support caching of values retrieves from Azure CLI to avoid unnecessary repeated invocations

manicminer:
This implements caching of important primitives: `AzVersion`, `DefaultSubscriptionId`, and `DefaultTenantId`. The first time these are needed by the authorizer, they are obtained by calling az-cli, but subsequent retrievals (e.g. from additional authorizers being set up) are from a package-level cache (in the `auth` package), yielding improved performance.

**Before**

SCR-20231128-pxw

**After**

SCR-20231128-pt6

I believe it's better to implement this within the authorizer, rather than at the lower-level `azurecli` internal helper package. With the latter, it'd be more robust as we could index the cache by the actual operation (caching the actual output) however that would probably need to be toggled at the package level which could lead to race conditions down the line if the `azurecli` package gets used anywhere else.

Resolves: #752
Replaces: #707
Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/23977

### DIFF
--- a/sdk/auth/azure_cli_authorizer.go
+++ b/sdk/auth/azure_cli_authorizer.go
@@ -17,6 +17,14 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type cachedCliData struct {
+	AzVersion             *string
+	DefaultSubscriptionId *string
+	DefaultTenantId       *string
+}
+
+var cliCache cachedCliData
+
 type AzureCliAuthorizerOptions struct {
 	// Api describes the Azure API being used
 	Api environments.Api
@@ -28,11 +36,15 @@ type AzureCliAuthorizerOptions struct {
 	// used for Resource Manager when auxiliary tenants are needed.
 	// e.g. https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant
 	AuxTenantIds []string
+
+	// EnableCaching permits module-scoped caching of the parsed results of `az` command invocations, and can be set to improve
+	// performance when instantiating multiple authorizers, at the cost of detectability during long-running applications.
+	EnableCaching bool
 }
 
 // NewAzureCliAuthorizer returns an Authorizer which authenticates using the Azure CLI.
 func NewAzureCliAuthorizer(ctx context.Context, options AzureCliAuthorizerOptions) (Authorizer, error) {
-	conf, err := newAzureCliConfig(options.Api, options.TenantId, options.AuxTenantIds)
+	conf, err := newAzureCliConfig(options.Api, options.TenantId, options.AuxTenantIds, options.EnableCaching)
 	if err != nil {
 		return nil, err
 	}
@@ -58,13 +70,23 @@ func (a *AzureCliAuthorizer) Token(_ context.Context, _ *http.Request) (*oauth2.
 		return nil, fmt.Errorf("could not request token: conf is nil")
 	}
 
-	azArgs := []string{"account", "get-access-token"}
+	var err error
 
 	// verify that the Azure CLI supports MSAL - ADAL is no longer supported
-	err := azurecli.CheckAzVersion(azurecli.MsalVersion, nil)
-	if err != nil {
+	var azVersion *string
+	if a.conf.CacheEnabled && cliCache.AzVersion != nil {
+		azVersion = cliCache.AzVersion
+	} else {
+		if azVersion, err = azurecli.GetAzVersion(); err != nil {
+			return nil, fmt.Errorf("%s. Please ensure you have installed Azure CLI version %s or newer", err, AzureCliMinimumVersion)
+		}
+	}
+	if err = azurecli.CheckAzVersion(*azVersion, azurecli.MsalVersion, nil); err != nil {
 		return nil, fmt.Errorf("checking the version of the Azure CLI: %+v", err)
 	}
+
+	azArgs := []string{"account", "get-access-token"}
+
 	scope, err := environments.Scope(a.conf.Api)
 	if err != nil {
 		return nil, fmt.Errorf("determining scope for %q: %+v", a.conf.Api.Name(), err)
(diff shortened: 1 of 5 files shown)


sdk/auth/azure_cli_authorizer.go
```diff
@@ -28,11 +36,15 @@ type AzureCliAuthorizerOptions struct {
 	// used for Resource Manager when auxiliary tenants are needed.
 	// e.g. https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/authenticate-multi-tenant
 	AuxTenantIds []string
+
+	// EnableCaching permits module-scoped caching of the parsed results of `az` command invocations, and can be set to improve
+	// performance when instantiating multiple authorizers, at the cost of detectability during long-running applications.
+	EnableCaching bool
```
tombuildsstuff, line 42: Is there a benefit to making caching configurable? This feels like something we'd always want on?

```suggestion
```

sdk/auth/azure_cli_authorizer.go
```diff
@@ -161,38 +191,75 @@ type azureCliConfig struct {
...
+		}
+
+		if cacheEnabled {
+			cliCache.AzVersion = azVersion
+		}
+	}
+	nextMajor := AzureCliNextMajorVersion
+	if err = azurecli.CheckAzVersion(*azVersion, azurecli.MsalVersion, &nextMajor); err != nil {
 		return nil, err
 	}
```
tombuildsstuff, line 220: If we make `cliCache` a pointer, we should be able to populate the entire object one-time, which'd achieve the same thing albeit a little simpler?

```suggestion
	if cliCache == nil {
		var err error
		azVersion, err = azurecli.GetAzVersion()
		if err != nil {
			return nil, fmt.Errorf("%s. Please ensure you have installed Azure CLI version %s or newer", err, AzureCliMinimumVersion)
		}
		var defaultTenantId *string
		if defaultTenantId, err = azurecli.CheckTenantID(tenantId); err != nil {
			return nil, err
		}
		if defaultTenantId == nil {
			return nil, errors.New("invalid tenantId or unable to determine tenantId")
		}
		var defaultSubscriptionId *string
		if defaultSubscriptionId, err = azurecli.GetDefaultSubscriptionID(); err != nil {
			return nil, err
		}
		cliCache = &cliCache{
			AzVersion: azVersion,
			DefaultSubscriptionId: defaultSubscriptionId,
			DefaultTenantId: defaultTenantId,
		}
	}
	
	nextMajor := AzureCliNextMajorVersion
	if err = azurecli.CheckAzVersion(*azVersion, azurecli.MsalVersion, &nextMajor); err != nil {
		return nil, err
	}
```

sdk/auth/config.go
```diff
@@ -21,6 +21,8 @@ type Credentials struct {
 
 	// EnableAuthenticatingUsingAzureCLI specifies whether Azure CLI authentication should be checked.
 	EnableAuthenticatingUsingAzureCLI bool
+	// AzureCLIUseCache specifies whether caching should be enabled for idempotent operation results
+	AzureCLIUseCache bool
```
tombuildsstuff, line 25: Which'd mean we could remove this?

```suggestion
```

tombuildsstuff (changes requested):
I've taken a look through and left some comments inline, but on the whole this change looks good, however I think we probably should look to:

1. Remove the `CachingEnabled` toggle - is there a use-case where this _wouldn't_ be beneficial? In the event there's an issue that arises from this, we can do another release of `hashicorp/go-azure-sdk` (which is more work, but we're going to want this cache enabled everywhere?)
2. Change how the cache is built, so that we're building this only when the object is nil (which achieves this one-time load without the `enabled` flag)
3. Centralise the Azure CLI checks, as we're obtaining details about the Azure CLI in `newAzureCliConfig` - meaning we don't need to re-check this in `Token` and `AuxiliaryTokens`

WDYT?

sdk/auth/azure_cli_authorizer.go
manicminer: Sounds good to me, I've refactored accordingly 👍

manicminer:
@tombuildsstuff I've refactored this so the caching now happens within the `azurecli` helper package. Instead of explicitly caching specific primitives, the cache is now automatically indexed by the `az` command arguments and populated with the resulting bytes. This improves the efficiency as now the same command can be re-used (as is the same for obtaining a default tenant ID and subscription ID) and achieves a cache hit every time. It also reduces complexity in the `auth` package, or anywhere else we might interface with az-cli. WDYT?

tombuildsstuff:
Totally missed this was sitting for a re-review, sorry 👀

sdk/internal/azurecli/azcli.go
```diff
@@ -59,68 +48,107 @@ func CheckAzVersion(minVersion string, nextMajorVersion *string) error {
...
 // CheckTenantID validates the supplied tenant ID, and tries to determine the default tenant if a valid one is not supplied.
-func CheckTenantID(tenantId string) (string, error) {
+func CheckTenantID(tenantId string) (*string, error) {
 	validTenantId, err := regexp.MatchString("^[a-zA-Z0-9._-]+$", tenantId)
 	if err != nil {
-		return "", fmt.Errorf("could not parse tenant ID %q: %s", tenantId, err)
+		return nil, fmt.Errorf("could not parse tenant ID %q: %s", tenantId, err)
 	}
 
 	if !validTenantId {
```
tombuildsstuff, line 58: _nit / not a blocker_ but since `GetDefaultSubscriptionID` exists perhaps this method wants splitting into `GetDefaultTenantID` and `CheckTenantID`?

tombuildsstuff (approved):
LGTM 👍

sdk/internal/azurecli/azcli.go
manicminer: done :)

orgads:
Thank you! Much better now.

PR #767, opened 2023-12-13
poller_lro: add pending status `CancelInProgress` for mysql

sinbai:
Fix the following acctest failure.

image


stephybun (approved):
Thanks @sinbai LGTM 👍

PR #772, opened 2023-12-13
automation: moving into the azurerm directory to determine if there's changes

tombuildsstuff:


tombuildsstuff:
Skipping review since this is only testable in GHA

PR #773, opened 2023-12-13
automation: writing the result to a file

tombuildsstuff:
Trying to make this more reproducible locally


tombuildsstuff:
Tested this locally but this can only really be tested end-to-end in GHA so skipping review

PR #781, opened 2023-12-14
environments: fix identifier URI for storage in china

manicminer:
Storage in china actually uses `https://storage.azure.com` as identifier URI, not `https://core.chinacloudapi.cn`

SCR-20231214-uiq

SCR-20231214-uhn


mbfrahry (approved):
LGTM!

PR #782, opened 2023-12-15
automation: outputting the value and not the key/value into the automation file

tombuildsstuff:
This file is used to output the variable for whether there's changes to push or not, but since we're outputting this into a variable elsewhere, we can just output this as the value and not a key/value pair here.


tombuildsstuff:
Skipping review since this is a minor nit that can only be tested in GHA

PR #793, opened 2024-01-02
do not pass --tenant argument for MIs using az-cli

harshavmb:
According to the documentation [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/azure_cli), `az login --identity` is a supported feature. Nevertheless, in current sdk, we are passing `--tenant` argument for all except `cloudshell`. `az-cli` doesn't seem to be accepting `--tenant` argument for managed identities. 

This has been flagged as an issue on `azurerm` provider [#24295](https://github.com/hashicorp/terraform-provider-azurerm/issues/24295). 

As far as this PR is concerned, we simply identify the MI by comparing `id` (subscriptionId) & `tenantId` fields returned as an output of `az account show` & for Managed Identities both are same. 

As an alternate, we could parse the output of above command & identify `user` type which have something like `userManagedIdentity` & `systemManagedIdentity` as `name` attribute followed by `servicePrincipal` as `type` but I think it's an overkill as by definition managed identities are at tenant level.

### DIFF
--- a/sdk/auth/azure_cli_authorizer.go
+++ b/sdk/auth/azure_cli_authorizer.go
@@ -65,8 +65,22 @@ func (a *AzureCliAuthorizer) Token(_ context.Context, _ *http.Request) (*oauth2.
 	}
 	azArgs = append(azArgs, "--scope", *scope)
 
+	defaultSubscriptionId, err := azurecli.GetDefaultSubscriptionID()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


manicminer:
Hi @harshavmb, thanks for opening this PR! In order to detect use of a managed identity with reasonable confidence, per your comment we will need to inspect the `user` object and make a best judgement on whether to omit the tenant argument. I've opened #910 to do this, which should resolve this issue with minimal chance of affecting other users. I'm going to go ahead and close this out this PR in favor of that one, thanks again for looking at this.

PR #819, opened 2024-01-18
dependencies: bumping `go-azure-helpers` to `v0.66.1`

stephybun:


stephybun:
Dependent on #818 and https://github.com/hashicorp/go-azure-helpers/pull/206

PR #834, opened 2024-01-23
automation: splitting the determining and publishing of the Git Tag

tombuildsstuff:
This is needed as a part of https://github.com/hashicorp/pandora/issues/3601 since we're going to be publishing multiple Go Modules at different paths - but the version number of each of them is going to be the same.

As such this PR splits the determining and publishing of the Git Tag - such that the existing update tooling continues to work with minor changes (tracked in https://github.com/hashicorp/terraform-provider-azurerm/pull/24602).

Given this is a Github Action this is hard to test locally - I've tested the scripts (which are fine) - but I suspect this change should be fine to merge today, prior to enabling module support.

### DIFF
--- a/.github/workflows/automation-release.yaml
+++ b/.github/workflows/automation-release.yaml
@@ -14,7 +14,7 @@ jobs:
     permissions:
       contents: write
     outputs:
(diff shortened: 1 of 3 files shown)


manicminer (approved):
LGTM 👍

PR #843, opened 2024-01-24
Go Modules Part 2 - `resource-manager/`

tombuildsstuff:
This PR is part 2 of adding Nested Go Modules - adding the `./resource-manager/` Go Module.

Once merged this likely wants to be manually released first - at which point we should be able to re-enable the automation.

Part of https://github.com/hashicorp/pandora/issues/3601

~**Note:** The Go Module path for the Base Layer needs updating.~ Done

~Dependent on https://github.com/hashicorp/go-azure-sdk/pull/842~


manicminer (approved):
pending the go.mod TODO, this LGTM 👍

PR #844, opened 2024-01-24
Go Modules Part 3 - `microsoft-graph/`

tombuildsstuff:
This PR is part 3 of adding Nested Go Modules - adding the `./microsoft-graph/` Go Module.

Once merged this likely wants to be manually released first - at which point we should be able to re-enable the automation.

At this point this Go Module is purely being scaffolded out / contains nothing usable - but will in the future - as such enabling automation here from the start allows us to use this as soon as it's available.

Part of https://github.com/hashicorp/pandora/issues/3601

Dependent on https://github.com/hashicorp/go-azure-sdk/pull/843


manicminer (approved):
LGTM 🍉

PR #849, opened 2024-01-25
automation: updating the go module update to use sed

tombuildsstuff:
Using `go get` on the SDK module means that we can end up unintentionally downgrading, since we're sure the SDK Version
exists (since we've just published it) - `sed` ensures this value will be what we're expecting.


manicminer (approved):
LGTM 🍟

PR #850, opened 2024-01-25
makefile: updating `prepare` to remove only directories

tombuildsstuff:
Removing the entire directory removes the `go.mod` and README etc, so needs updating

Fixes https://github.com/hashicorp/pandora/pull/3702 and https://github.com/hashicorp/pandora/pull/3703


manicminer (approved):
LGTM 👍

PR #879, opened 2024-02-09
sdk: check for empty response body before unmarshalling

catriona-m:
Fixes a bug with Maintenance Delete where the response body is empty, but not nil, so this does not get unmarshalled.

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -189,6 +189,11 @@ func (r *Response) Unmarshal(model interface{}) error {
 		// Trim away a BOM if present
 		respBody = bytes.TrimPrefix(respBody, []byte("\xef\xbb\xbf"))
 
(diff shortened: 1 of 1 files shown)


sdk/client/client.go
```diff
@@ -189,6 +189,11 @@ func (r *Response) Unmarshal(model interface{}) error {
 		// Trim away a BOM if present
 		respBody = bytes.TrimPrefix(respBody, []byte("\xef\xbb\xbf"))
 
+		// In some cases the respBody is empty, but not nil, so don't attempt to unmarshal this
+		if len(respBody) == 0 {
+			return nil
+		}
```
tombuildsstuff, line 195: Mind implementing the same fix for XML deserialization too (but _not_ for the bytes deserialization, which'd want that context)?

tombuildsstuff (approved):
One comment that we should fix this for XML too, but otherwise 👍

PR #902, opened 2024-02-27
bugfix: include the xml doctype declaration when marshaling

manicminer:


sdk/client/client.go
```diff
@@ -104,6 +104,9 @@ func (r *Request) Marshal(payload interface{}) error {
 	case strings.Contains(contentType, "application/xml") || strings.Contains(contentType, "text/xml"):
 		body, err := xml.Marshal(payload)
 		if err == nil {
+			// Prepend the xml doctype declaration since it is not automatically added
+			body = append([]byte(xml.Header), body...)
```
tombuildsstuff, line 108: This might be problematic in either `Automation` or `APIM` (I forget which) where the (user specified) body can include an xml doctype - might be worth a `startswith` check until that's updated to account for this?

tombuildsstuff (approved):
One comment about a usage which might be problematic but otherwise 👍

sdk/client/client.go
manicminer: Makes sense, will do 👍

PR #927, opened 2024-03-14
Odata: enhance nextLink unmarshal

teowa:
try to resolves https://github.com/hashicorp/terraform-provider-azurerm/issues/24948

Seems the error is because code-gen does not handle [`x-ms-pageable.nextLinkName`](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/creating-swagger.md#paging-with-x-ms-pageable)

And the `nextLink` is used in https://github.com/hashicorp/go-azure-sdk/blob/8ee56b008ba9275ff6c3282052e6b50e83073f9a/sdk/client/client.go#L538-L579

This PR simply hardcode the `nextLink` to handle the case pageable List API returns next link through `nextLink` field instead of standard `@odata.nextLink` field. 

A more robust solution is to make use of [`x-ms-pageable.nextLinkName`](https://github.com/Azure/azure-rest-api-specs/blob/main/documentation/creating-swagger.md#paging-with-x-ms-pageable) for pageable List API at the [generator-go-sdk](https://github.com/hashicorp/pandora/blob/368ec110f6eb14be8bea9a55357770b797739cb5/tools/generator-go-sdk/internal/generator/templater_methods.go#L312) , but it might break existing code struct.

### DIFF
--- a/sdk/odata/odata.go
+++ b/sdk/odata/odata.go
@@ -93,14 +93,14 @@ type OData struct {
 	MetadataEtag *string `json:"@odata.metadataEtag"`
 	Type         *Type   `json:"@odata.type"`
 	Count        *int    `json:"@odata.count"`
-	NextLink     *Link   `json:"@odata.nextLink"`
 	Delta        *string `json:"@odata.delta"`
 	DeltaLink    *Link   `json:"@odata.deltaLink"`
 	Id           *Id     `json:"@odata.id"`
 	EditLink     *Link   `json:"@odata.editLink"`
 	Etag         *string `json:"@odata.etag"`
 
-	Error *Error `json:"-"`
+	Error    *Error `json:"-"`
+	NextLink *Link  `json:"-"`
 
 	Value interface{} `json:"value"`
 }
@@ -127,6 +127,18 @@ func (o *OData) UnmarshalJSON(data []byte) error {
 				return err
 			}
 			o.Error = &e2
+			break
+		}
+	}
+
+	for _, k := range []string{"nextLink", "@odata.nextlink"} {
+		if v, ok := e[k]; ok {
+			var l2 Link
+			if err := json.Unmarshal(v, &l2); err != nil {
+				return err
+			}
+
+			o.NextLink = &l2
 			break
 		}
 	}


sdk/odata/odata.go
```diff
@@ -131,5 +131,17 @@ func (o *OData) UnmarshalJSON(data []byte) error {
 		}
 	}
 
+	for _, k := range []string{"nextLink", "@odata.nextlink"} {
```
tombuildsstuff, line 134: There's a number of possible values for this field depending on the part of the API in question, so these two fields aren't necessarily sufficient.

Instead the SDK supports [defining a custom pager](https://github.com/hashicorp/go-azure-sdk/blob/8ee56b008ba9275ff6c3282052e6b50e83073f9a/sdk/client/request_options.go#L25-L27) which, given that Pandora has the JSONName for the field containing the pagination details, would allow us to output a `CustomPager` ([like so](https://github.com/hashicorp/go-azure-sdk/blob/8ee56b008ba9275ff6c3282052e6b50e83073f9a/sdk/client/client_test.go#L161-L178)) which is then configured [on the `client.RequestObject` struct](https://github.com/hashicorp/go-azure-sdk/blob/8ee56b008ba9275ff6c3282052e6b50e83073f9a/sdk/client/request_options.go#L25-L27) and ultimately used [when the SDK method definition is output](https://github.com/hashicorp/go-azure-sdk/blob/8ee56b008ba9275ff6c3282052e6b50e83073f9a/resource-manager/apimanagement/2023-05-01-preview/apiissue/method_get.go#L50-L58).

We're currently using this information [to build out the AutoRest methods that use polling](https://github.com/hashicorp/pandora/blob/7b47fa7e752c27291ec8cb983e4bfb3005a31a5a/tools/generator-go-sdk/internal/generator/templater_methods_autorest.go#L610) so once the `CustomPoller` is output, that should be all that's needed.

As such, rather than doing this here - can we update [`generator-go-sdk` to output a `CustomPoller` as needed here](https://github.com/hashicorp/pandora/blob/7b47fa7e752c27291ec8cb983e4bfb3005a31a5a/tools/generator-go-sdk/internal/generator/templater_methods.go#L505-L514) instead, which'll fix this regardless of the field type?

tombuildsstuff (changes requested):
hey @teowa 

Thanks for this PR.

I've taken a look through and left a comment inline, and whilst this approach will work for the default values for ARM, the `nextLink` value can come from one of a number of fields depending on the API in question - as such I believe it'd be better to fix this in `hashicorp/pandora`'s `generator-go-sdk` tool, which has access to the name of the field containing the paginated value instead?

Thanks!

teowa:
Thanks @tombuildsstuff  for reviewing this and providing helpful detailed advice, I have submitted https://github.com/hashicorp/pandora/pull/3958 to solve the issue.

PR #933, opened 2024-03-19
enhancement: include HTTP status text in error when response is unexpected

manicminer:
Minor improvement, but adds a small amount of context mainly useful in the case that the API returns no error text.

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -513,10 +513,21 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 
 	// Determine whether response status is valid
 	if !containsStatusCode(req.ValidStatusCodes, resp.StatusCode) {
-		// The status code didn't match, but we also need to check the ValidStatusFUnc, if provided
+		// The status code didn't match, but we also need to check the ValidStatusFunc, if provided
 		// Note that the odata argument here is a best-effort and may be nil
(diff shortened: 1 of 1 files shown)


sdk/client/client.go
```diff
@@ -513,12 +513,17 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
...
 	if !containsStatusCode(req.ValidStatusCodes, resp.StatusCode) {
-		// The status code didn't match, but we also need to check the ValidStatusFUnc, if provided
+		// The status code didn't match, but we also need to check the ValidStatusFunc, if provided
 		// Note that the odata argument here is a best-effort and may be nil
 		if f := req.ValidStatusFunc; f != nil && f(resp.Response, resp.OData) {
 			return resp, nil
 		}
 
+		status := fmt.Sprintf("%d", resp.StatusCode)
+		if statusText := http.StatusText(resp.StatusCode); statusText != "" {
```
tombuildsstuff, line 523: Azure doesn't always use the Status Code to mean what the HTTP status means, but the HTTP Status Text is returned along side this, so we should be able to make this:

```suggestion
		if statusText := resp.Status; statusText != nil {
```

tombuildsstuff (approved):
One comment but otherwise LGTM 👍

tombuildsstuff (approved):
LGTM 👍

PR #945, opened 2024-03-26
Extend retryablehttp limits in SDK base client

manicminer:
Explicitly set three fields in the `retryablehttp.Client` struct.

`RetryWaitMin` is the shortest wait time before attempting to retry the request, and is the same as the default, but set explicitly here for clarity.

`RetryWaitMax` is the longest wait time before attempting to retry the request. The default is 30 seconds, we extend this to 61 second in consideration of potential API throttling for especially long-lasting eventual consistency issues.

With these values, the retry curve looks something [like this](https://go.dev/play/p/jlW1Lwc9VNT).

`RetryMax` is the maximum number of retries for a given API request, and is given the most consideration here. The default is 4, which is far too little for our purposes. Here, we set it to 16, which in conjunction with the above two field values, and when implemented using an exponential backoff algorithm, should take approximately 10 minutes to exhaust.

However, in consideration of _really_ long lasting retries, we inspect the context and increase this if it looks like the retries will be exhausted before the context deadlines. The goal is to try and deadline the context just before the retries are exhausted.

For compatibility, since we don't strictly require a deadline to execute a request, we don't require one here.

Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/25397

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -436,7 +436,7 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 	}
 
 	// Instantiate a RetryableHttp client and configure its CheckRetry func
-	r := c.retryableClient(func(ctx context.Context, r *http.Response, err error) (bool, error) {
+	r := c.retryableClient(ctx, func(ctx context.Context, r *http.Response, err error) (bool, error) {
 		// First check for badly malformed responses
 		if r == nil {
 			if req.IsIdempotent() {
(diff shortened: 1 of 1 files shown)


sdk/client/client.go
```diff
@@ -672,6 +672,18 @@ func (c *Client) retryableClient(checkRetry retryablehttp.CheckRetry) (r *retrya
 	r.CheckRetry = checkRetry
 	r.ErrorHandler = RetryableErrorHandler
 	r.Logger = log.Default()
+	r.RetryWaitMin = 1 * time.Second
+	r.RetryWaitMax = 61 * time.Second
```
tombuildsstuff, line 676: FWIW some [ARM API's do have `Retry-After` headers with an excessive timeout (e.g. 10m)](https://github.com/Azure/azure-rest-api-specs/issues/13115) - but I don't see any harm in us polling more regularly given that extended interval so 👍

tombuildsstuff (approved):
One comment but otherwise LGTM 👍

sdk/client/client.go
manicminer: Oh if the response includes a `Retry-After` header, that is always selected first before falling back to exponential backoff (i.e. these are not used in that case)

PR #960, opened 2024-04-09
Metadata fixes

manicminer:
> [!WARNING]  
> This must be merged in tandem with https://github.com/hashicorp/terraform-provider-azurerm/pull/25546, to avoid breaking `storage`

- environments: fix up refreshing from metadata, add missing endpoints/resource IDs for public, usgov and china
- metadata: remove "support" for metadata 2019-05-01 since it does not include MS Graph and will no longer work with Terraform providers anyway


sdk/environments/environment.go
```diff
@@ -160,7 +160,7 @@ func NewApiEndpoint(name, endpoint string, appId *string) *ApiEndpoint {
 	}
 }
 
-func (e *ApiEndpoint) withResourceIdentifier(identifier string) *ApiEndpoint {
+func (e *ApiEndpoint) WithResourceIdentifier(identifier string) Api {
 	e.resourceIdentifier = pointer.To(identifier)
 	return e
 }
```
tombuildsstuff, line 166: Since there's talk of overriding the token below on a per-Storage Account basis - perhaps this should return a new instance of `Api` rather than mutating the existing object, to ensure this is a clean-copy?

tombuildsstuff (approved):
A few minor things but otherwise LGTM 👍

PR #970, opened 2024-04-19
sdk/client: conditionally return the error when the resp is nil whilst retrying the request

manicminer:
ref: https://github.com/manicminer/hamilton/pull/280


catriona-m (approved):
LGTM!

PR #1008, opened 2024-06-03
auth: support subscription hint for Azure CLI, to help with multiple signed-in accounts with segregated access

manicminer:
Enables Azure CLI authentication to work in the case that multiple accounts are signed in, and each account has access to a subset of the available subscriptions. Currently, we pass the tenant ID to authenticate against which leaves Azure CLI with a decision to make as for which account to obtain an access token. Usually it will look at the default account (via `az account set`) and if this matches the tenant, it'll use that. If it doesn't match, it's forced to pick one of the accounts to use. Either of these choices may not align with the Terraform config, especially if it contains aliased providers using different subscriptions.

By passing the subscription ID, we inform it (when the API being authorized is Resource Manager) of the subscription we're going to be using, so it can make a better judgement about the account to select based on the subscription(s) that account has permissions for. This may fall down if one account has more limited permissions for a given subscription than another account, but that is left to the user to assign appropriate roles as needed.

Related: https://github.com/hashicorp/terraform-provider-azurerm/issues/23068

### DIFF
--- a/sdk/auth/auth_test.go
+++ b/sdk/auth/auth_test.go
@@ -183,6 +183,7 @@ func TestAccNewAuthorizerFromCredentials(t *testing.T) {
 
 	credentials := auth.Credentials{
 		AuxiliaryTenantIDs:            test.AuxiliaryTenantIds,
+		AzureCliSubscriptionIDHint:    test.SubscriptionId,
 		ClientCertificateData:         test.Base64DecodeCertificate(t, test.ClientCertificate),
 		ClientCertificatePassword:     test.ClientCertPassword,
 		ClientCertificatePath:         test.ClientCertificatePath,
--- a/sdk/auth/auth_test.go
+++ b/sdk/auth/auth_test.go
@@ -183,7 +183,6 @@ func TestAccNewAuthorizerFromCredentials(t *testing.T) {
 
 	credentials := auth.Credentials{
 		AuxiliaryTenantIDs:            test.AuxiliaryTenantIds,
-		AzureCliSubscriptionIDHint:    test.SubscriptionId,
 		ClientCertificateData:         test.Base64DecodeCertificate(t, test.ClientCertificate),
 		ClientCertificatePassword:     test.ClientCertPassword,
 		ClientCertificatePath:         test.ClientCertificatePath,
@@ -206,6 +205,7 @@ func TestAccNewAuthorizerFromCredentials(t *testing.T) {
 			credentials: func() (ret auth.Credentials) {
 				ret = credentials
 				ret.EnableAuthenticatingUsingAzureCLI = true
+				ret.AzureCliSubscriptionIDHint = test.SubscriptionId
 				return
 			},
 			check: func(a auth.Authorizer) error {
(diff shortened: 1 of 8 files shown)


sdk/auth/azure_cli_authorizer.go
```diff
@@ -199,11 +218,34 @@ func newAzureCliConfig(api environments.Api, tenantId string, auxiliaryTenantIds
...
+			}
+		}
+	}
+
 	return &azureCliConfig{
 		Api:                   api,
 		TenantID:              tenantId,
 		AuxiliaryTenantIDs:    auxiliaryTenantIds,
 		DefaultSubscriptionID: subscriptionId,
+		SubscriptionIDHint:    strings.ToLower(subscriptionIdHint),
```
tombuildsstuff, line 248: lower-casing the Subscription ID _feels_ like it'd be potentially problematic?

manicminer: I think we should move to normalize this wherever possible, although az-cli and the API accept any casing here. In this case there's no knock-on effect since this field isn't consumed/referenced anywhere, except when acquiring an access token.

sdk/auth/auth_test.go
```diff
@@ -183,6 +183,7 @@ func TestAccNewAuthorizerFromCredentials(t *testing.T) {
 
 	credentials := auth.Credentials{
 		AuxiliaryTenantIDs:            test.AuxiliaryTenantIds,
+		AzureCliSubscriptionIDHint:    test.SubscriptionId,
```
tombuildsstuff, line 186: for this one, do we only want to set this for the CLI test-case below? (207/208)?

sdk/auth/azure_cli_authorizer.go
tombuildsstuff: 👍 as discussed offline, this instance is fine - we'll want to take this into account elsewhere too

sdk/auth/config.go
```diff
@@ -21,6 +21,7 @@ type Credentials struct {
 
 	// EnableAuthenticatingUsingAzureCLI specifies whether Azure CLI authentication should be checked.
 	EnableAuthenticatingUsingAzureCLI bool
+	AzureCliSubscriptionIDHint        string
```
tombuildsstuff, line 24: docs/ordering?

sdk/environments/interfaces.go
```diff
@@ -26,3 +30,18 @@ type Api interface {
 	// WithResourceIdentifier overrides the default resource ID for the API and is useful for APIs that offer multiple authorization scopes
 	WithResourceIdentifier(string) Api
 }
+
+func ApiIsKnown(api Api, apiName string) bool {
```
tombuildsstuff, line 34: docs?

tombuildsstuff (approved):
hey @manicminer 

On the whole this is looking pretty good, I've left a few minor comments inline but this otherwise LGTM 👍

sdk/auth/config.go
manicminer: oops, adding 😁

PR #1010, opened 2024-06-04
`sdk/client/resourcemanager`: updating the `LRO` poller to treat a 404 as `InProgress` / tests covering the `LRO`/`ProvisioningState` polling scenarios

tombuildsstuff:
This PR updates the `LRO` poller to treat 404's as `InProgress`, which should fix https://github.com/hashicorp/go-azure-sdk/issues/989

Whilst threading this through I've also added some tests which make use of a fake HTTP server to both more realistically test both the LRO and ProvisioningState poller - but also (hopefully) better explain the scenarios to other contributors.

### DIFF
--- a/sdk/client/resourcemanager/poller_helpers_test.go
+++ b/sdk/client/resourcemanager/poller_helpers_test.go
@@ -1,0 +1,57 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resourcemanager
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/hashicorp/go-azure-helpers/lang/pointer"
+)
+
+type expectedResponse struct {
+	dropsConnection      bool
+	httpStatusCode       *int
+	status               *status
+	useStatus            bool
+	useProvisioningState bool
+}
+
+func responseThatDropsTheConnection() expectedResponse {
+	return expectedResponse{
+		dropsConnection: true,
+	}
+}
+
+func responseWithStatusInProvisioningState(input status) expectedResponse {
+	return expectedResponse{
+		status:               pointer.To(input),
+		useProvisioningState: true,
+	}
+}
+
+func responseWithStatusInStatusField(input status) expectedResponse {
+	return expectedResponse{
+		status:    pointer.To(input),
+		useStatus: true,
+	}
+}
+
+func responseWithHttpStatusCode(input int) expectedResponse {
+	return expectedResponse{
+		httpStatusCode: pointer.To(input),
+	}
+}
+
+func dropConnection(t *testing.T, w http.ResponseWriter) {
+	httpHijacker, ok := w.(http.Hijacker)
+	if !ok {
+		t.Fatalf("`w` didn't implement `http.httpHijacker`")
+	}
+	conn, _, err := httpHijacker.Hijack()
+	if err != nil {
+		return
+	}
+	conn.Close()
+}
--- a/sdk/client/resourcemanager/poller_provisioning_state_helpers_test.go
+++ b/sdk/client/resourcemanager/poller_provisioning_state_helpers_test.go
@@ -94,6 +94,9 @@ func (e provisioningStateEndpoint) assertCalled(t *testing.T, expected int) {
 func (e provisioningStateEndpoint) response() *http.Response {
 	return &http.Response{
 		Header: map[string][]string{
+			// NOTE: any custom host/port is discarded, since we use the host from the Client when building the
+			// Poller - this is because we should only be hitting the ARM Endpoint (and using that token) so
+			// any differences (including regional endpoints, or custom ports) would be unexpected/an error.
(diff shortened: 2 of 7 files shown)


sdk/client/resourcemanager/poller_helpers_test.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package resourcemanager
```
manicminer, line 4: Would it be better to use a test package for these? I might have missed something but it doesn't seem like we're testing any unexported functions?

sdk/client/resourcemanager/poller_provisioning_state_helpers_test.go
```diff
@@ -0,0 +1,101 @@
...
+		if apiv := r.URL.Query().Get("api-version"); apiv != e.expectedApiVersion {
+			w.WriteHeader(http.StatusNotImplemented)
+			t.Fatalf("expected the `api-version` to be set to `%s` but got %q", e.expectedApiVersion, apiv)
+			return
+		}
+
+		e.numberOfTimesCalled++
+		if e.numberOfTimesCalled > len(e.responses) {
+			t.Fatal("called too many times")
+		}
```
manicminer, line 53: Could we just return the last `expectedResponse` when the `responses` slice is overrun?

sdk/client/resourcemanager/poller_provisioning_state_helpers_test.go
```diff
@@ -0,0 +1,101 @@
...
+func (e provisioningStateEndpoint) assertCalled(t *testing.T, expected int) {
+	if e.numberOfTimesCalled != expected {
+		t.Fatalf("expected to be called %d times but got %d", expected, e.numberOfTimesCalled)
+	}
+}
+
+func (e provisioningStateEndpoint) response() *http.Response {
+	return &http.Response{
+		Header: map[string][]string{
+			http.CanonicalHeaderKey("Azure-AsyncOperation"): {"http://localhost/lro/poll"},
```
manicminer, line 97: Might be worth a comment that this relies on the poller dropping the host/port from the URI and using the API's base URI?

sdk/client/resourcemanager/poller_provisioning_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// Ensure we're parsing from the `provisioningState` field
+	helper := newProvisioningStateEndpoint([]expectedResponse{
+		responseWithStatusInProvisioningState(statusSucceeded),
+	})
+	server := httptest.NewServer(http.HandlerFunc(helper.endpoint(t)))
+	defer server.Close()
+
+	resourceManagerClient := &Client{
+		// NOTE: the use of a different API Version here is _intentional_ to ensure it's unused since the
```
manicminer, line 29: Comment looks accidentally curtailed? :)

sdk/client/resourcemanager/poller_provisioning_state_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// Ensure we're parsing from the `status` field
+	helper := newProvisioningStateEndpoint([]expectedResponse{
+		responseWithStatusInStatusField(statusSucceeded),
+	})
+	server := httptest.NewServer(http.HandlerFunc(helper.endpoint(t)))
+	defer server.Close()
+
+	resourceManagerClient := &Client{
+		// NOTE: the use of a different API Version here is _intentional_ to ensure it's unused since the
```
manicminer, line 61: (as above)

manicminer (approved):
Thanks @tombuildsstuff, this looks great. Just a few queries but otherwise LGTM 👍

sdk/client/resourcemanager/poller_helpers_test.go
tombuildsstuff: Possibly, I don't think we particularly gain anything from doing that here though, so I'd question the need for that?

sdk/client/resourcemanager/poller_provisioning_state_helpers_test.go
tombuildsstuff: We could, but since we're testing that the specific requests/responses are returning what we're expecting, I think that'd be potentially misleading in the test helper, so I'd rather not do that for now until there's a compelling use-case?

manicminer: I just thought it might more closely model the API behavior? i.e. when a final state is reached, you'd expect to see that in subsequent responses

tombuildsstuff: Depends on the API implementation - in some cases that's true (and it remains available for up to 24h, as per the ARM spec) - in some other cases it's gone once it returns `Succeeded` one-time

manicminer: Ah ok, I didn't realise that 👍

PR #1046, opened 2024-07-11
Fix Missing AppId that I missed earlier, and add MicrosoftIntune 0000…

gbonk:
…000a-0000-0000-c000-000000000000


manicminer (approved):
Thanks @gbonk, this LGTM 👍 

These Intune client IDs are amusing...

PR #1051, opened 2024-07-15
Add AwaitingManualAction status

chatlabashev:
Adding AwaitingManualAction status, which is returned when we are creating an Event subscription with manual webhook validation

polling after CreateOrUpdate: result.Status was nil/empty - op.Status was "AwaitingManualAction" / op.Properties.ProvisioningState was ""


chatlabashev:
This is to remediate the error seem in this issue:
https://github.com/hashicorp/terraform-provider-azurerm/issues/25689

manicminer (approved):
Thanks @chatlabashev, this LGTM 👍

PR #1067, opened 2024-08-28
`sdk/client/resourcemanager`: supporting `Building` as an undocumented polling status

catriona-m:


stephybun (approved):
Thanks @catriona-m LGTM 👍

PR #1079, opened 2024-09-13
Update types.go

k3v3n:
self explanatory. 

- https://learn.microsoft.com/en-us/graph/api/resources/internaldomainfederation
- https://learn.microsoft.com/en-us/graph/api/resources/authorizationpolicy


k3v3n:
Thanks for letting me know.

PR #1087, opened 2024-10-01
`SDK` - Limit retries for network level failures

jackofallops:
* add support for limiting retries for network level failures such as tcp dial timeouts

This PR updates handling of retries for network failures that are unlikely to recover in a reasonable timeframe (for example the client being behind a firewall silently dropping packets, or an un-routable address), specifically on `GET`, `HEAD` and `OPTIONS` requests, which would otherwise repeat endlessly until a parent context expired.


sdk/client/client.go
```diff
@@ -749,3 +751,95 @@ func containsStatusCode(expected []int, actual int) bool {
 
 	return false
 }
+
+// extendedRetryPolicy extends the defaultRetryPolicy implementation in go-retryablehhtp with
+// additional error conditions that should not be retried indefinitely
+// `tcpDialTimeoutRe` - covers the
```
stephybun, line 757: This comment looks incomplete?

stephybun (approved):
Thanks @jackofallops LGTM 🦭

Issue #962, opened 2024-04-12
Issue with [ApiManagement] / API Version [2022-08-01] - resourcemanager/pollers: LRO URIs of themselves should not be judged entirely based on Path


sinbai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Community Note

* Please vote on this issue by adding a :thumbsup: [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Service Used

ApiManagement

### API Versions Used

2022-08-01

### Description

The [code](https://github.com/hashicorp/go-azure-sdk/blob/166d9db9db0413fb1a95672edfc53003ed8e1b3d/sdk/client/resourcemanager/poller.go#L85) in `func isLROSelfReference(lroPollingUri, originalRequestUri string) bool `uses `Path` to determine whether lroPollingUri is itself, which does not apply to the following situation.

Take creating `azurerm_api_management_api` with an invalid open api file as an example.

TF config:
```
terraform {
  required_providers {
    azurerm = {
      source = "hashicorp/azurerm"
      version = "3.98.0"
    }
  }
}


provider "azurerm" {
  features {}
}

resource "azurerm_resource_group" "test" {
  name     = "exampleRG"
  location = "eastus2"
}

resource "azurerm_api_management" "test" {
  name                = "example"
  location            = azurerm_resource_group.test.location
  resource_group_name = azurerm_resource_group.test.name
  publisher_name      = "pub1"
  publisher_email     = "pub1@email.com"

  sku_name = "Consumption_0"
}

resource "azurerm_api_management_api" "test" {
  name                = "example"
  resource_group_name = azurerm_resource_group.test.name
  api_management_name = azurerm_api_management.test.name
  display_name        = "api1"
  path                = "api1"
  protocols           = ["https"]

  revision            = "1"
  
  import {
    content_value  = file("api_management_api_openapi.yaml")
    content_format = "openapi"
  }
}
```

api_management_api_openapi.yam
```
openapi: 3.0.1
info:
  title: j7t-reproducer-service
  version: "1"
servers:
- url: rest
  variables: {}
tags:
- name: tag
  description: description of tag
paths:
  /document:
    get:
      description: description of operation2
[...]

PR #1090, opened 2024-10-09
`sdk`: provisioning state poller allow additional parameters

wuxu92:
This PR is to fix issues that provisioning state poller may need additional query parameters to request the resource.

Possibly unblocks #962, https://github.com/hashicorp/terraform-provider-azurerm/issues/23322 and https://github.com/hashicorp/terraform-provider-azurerm/issues/24379

## Issue description
For the APIM API resource, the response to a create/update operation includes a `location` header with an `asyncId` parameter. This `asyncId` is needed to retrieve the actual error message if the create/update operation fails.

image

### Create
If missing the asyncId parameter, the GET operation after a failed creation will return 404 Not Found, obscuring the actual ValidationError.

image

When this PR is applied, the error message will reflect the root cause as follows:

image

### Update
The GET operation after a failed update will return 200 OK, mistakenly indicating the update was successful. More details can be found at https://github.com/hashicorp/terraform-provider-azurerm/issues/24379.

When this PR is applied, the update operation will report the error if the swagger contains validation error:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .fleet/
 .idea/
+.vscode/
 .DS_Store
 tmp/
 vendor/
--- a/sdk/client/resourcemanager/poller_provisioning_state.go
+++ b/sdk/client/resourcemanager/poller_provisioning_state.go
@@ -48,6 +48,11 @@ func provisioningStatePollerFromResponse(response *client.Response, lroIsSelfRef
 	}
 
 	resourcePath := originalUri
+	if pollingUriStr := pollingUriForLongRunningOperation(response); pollingUriStr != "" {
+		if pollingUri, err := url.ParseRequestURI(pollingUriStr); err == nil {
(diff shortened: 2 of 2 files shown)


sinbai:
I have tested this PR, it fixes the issue mentioned in #23322 (the case where the imported file contains errors), terraform returns the error information returned by the API instead of a 404 error. Details are as follows.

Before fix:
image

After fix:
image

jackofallops:
Hi @wuxu92 - Are there other known provisioning cases requiring this modification? I'm just wondering why we're making the change here, rather than creating a custom poller for this as we have done for other unique cases?

jackofallops:
> I created customr poller PR [hashicorp/terraform-provider-azurerm#28193](https://github.com/hashicorp/terraform-provider-azurerm/pull/28193)

Thanks @wuxu92 - I've reviewed the linked PR, and since we have away forward there I'm going to close this one out for now.

PR #1104, opened 2024-10-18
handle empty `nextlink` value as end of pagination

ziyeqf:
The `chaos` return an empty string in the `NextLink` property instead of set it to `null` when calling [`TargetTypes_List`
](https://github.com/Azure/azure-rest-api-specs/blob/8ad58020cfafc5ace71d85b31272acf4d6b694d3/specification/chaos/resource-manager/Microsoft.Chaos/stable/2023-11-01/targetTypes.json#L43C25-L43C41). It will make the provider wait endless...

I have opened an [issue](https://github.com/Azure/azure-rest-api-specs/issues/31098) on azure-rest-api-specs repo, but can we consider add this workaround to avoid similar bad cases?


jackofallops (approved):
Thanks @ziyeqf - Sorry for the delay, lots of moving parts and PTO on the team held this one up.  This LGTM 👍

PR #1108, opened 2024-10-23
Error Handling - update tcp dial timeout regex to catch broader range of error conditions

jackofallops:


sdk/client/resourcemanager/poller_provisioning_state.go
```diff
@@ -84,6 +89,17 @@ func (p *provisioningStatePoller) Poll(ctx context.Context) (*pollers.PollResult
...
+		if errors.As(err, &e) {
+			p.droppedConnectionCount++
+			if p.droppedConnectionCount < p.maxDroppedConnections {
+				return &pollers.PollResult{
+					PollInterval: p.initialRetryDuration,
+					Status:       pollers.PollingStatusUnknown,
+				}, nil
+			}
+		}
+
```
stephybun, line 102: Why don't we want to reset the count for dropped connections to 0 here like we do for the LRO poller?

stephybun (approved):
Thanks @jackofallops one question other LGTM 🦒

sdk/client/resourcemanager/poller_provisioning_state.go
jackofallops: Good spot! Thanks!

PR #1123, opened 2024-11-25
exclude Resource Manager endpoints from network failure detection

jackofallops:
Users on unstable/unreliable networks can experience terraform operation disruption. Since the extendedRetryPolicy is needed only for Data Plane clients, this PR restores previous "Retry Forever" (specifically until operation context deadline expires) behaviour for Resource Manager  endpoints.

* bump go to `1.22`
* exclude azure management endpoints from network failure logic.

### DIFF
--- a/sdk/client/client.go
+++ b/sdk/client/client.go
@@ -471,8 +471,13 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 		// Check for failed connections etc and decide if retries are appropriate
 		if r == nil {
 			if req.IsIdempotent() {
-				return extendedRetryPolicy(r, err)
+				if !isResourceMangerHost(req) {
+					return extendedRetryPolicy(r, err)
+				}
+
+				return retryablehttp.DefaultRetryPolicy(ctx, r, err)
 			}
+
 			return false, fmt.Errorf("HTTP response was nil; connection may have been reset")
 		}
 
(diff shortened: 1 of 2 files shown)


sdk/client/client.go
```diff
@@ -471,8 +471,13 @@ func (c *Client) Execute(ctx context.Context, req *Request) (*Response, error) {
 		// Check for failed connections etc and decide if retries are appropriate
 		if r == nil {
 			if req.IsIdempotent() {
-				return extendedRetryPolicy(r, err)
+				if !isResourceMangerHost(req) {
```
stephybun, line 474:
```suggestion
				if !isResourceManagerHost(req) {
```

sdk/client/client.go
```diff
@@ -840,3 +846,22 @@ func extendedRetryPolicy(resp *http.Response, err error) (bool, error) {
 
 	return false, nil
 }
+
+// exclude Resource Manager calls from the network failure retry avoidance to help users on unreliable networks
+// This code path should be removed when the Data Plane separation work is completed and 5.0 ships.
+func isResourceMangerHost(req *Request) bool {
```
stephybun, line 852:
```suggestion
func isResourceManagerHost(req *Request) bool {
```

stephybun (approved):
Minor typo, otherwise LGTM 🦚

PR #1127, opened 2024-11-26
Update automation-release.yaml

randyhdev:
The proposed changes add the service user token to actions checkout. Alternatively, we could disable persistence and set this in each step as needed.


stephybun (approved):
Thanks @randyhdev LGTM ❄️

PR #1138, opened 2024-12-31
[CODEOWNERS] - add ADR required reviewers for DeviceRegistry directory to CODEOWNERS

mryanlo:
Adding ADR required reviewers to the CODEOWNERS file so that any updates to the golang hashicorp azure sdk will be flagged and we can be aware of the changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,4 @@
 *  @hashicorp/terraform-azure
 
 # PRLabel: %Device Registry
-/resource-manager/deviceregistry/ @marcodalessandro @rohankhandelwal @riteshrao
+/resource-manager/deviceregistry/ @marcodalessandro @rohankhandelwal @riteshrao @davidemontanari
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,4 @@
 *  @hashicorp/terraform-azure
+
+# PRLabel: %Device Registry
+/resource-manager/deviceregistry/ @marcodalessandro @rohankhandelwal @riteshrao


mryanlo:
@wuxu92 could I please get a review for this PR? it's been open for a while now

mryanlo:
> @mryanlo I believe there might be some confusion; I don't have context regarding this Pull Request. From my understanding, the code owner isn't required here. Perhaps @stephybun can provide some suggestions.

Hi @wuxu92, the context is that we (microsoft's Azure Device Registry team) want the abillity to be tagged in any pull requests that make changes to the `deviceregistry` folder by updating the CODEOWNERS file. We have had regressions in other github repos for our services because we were not informed of the PR changes. This isn't a regression but an example of a change we were not aware of until later was this PR: https://github.com/hashicorp/go-azure-sdk/pull/1102 which added a new api version to the device registry service to be auto generated for the go lang sdk. This change is fine, but we'd like to know about it in the future, so the accounts in the code owners change I made will be notified.

mryanlo:
> Hi @mryanlo, Thank you for providing the context. There's a mechanism in place for automatically updating the go-azure-sdk with new stable API versions. Thus, you can anticipate the go-azure-sdk being updated some time after the API version is released in [Azure/azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs) . For reference, you can view an example of an automatic pull request to bump up the API version here: [hashicorp/pandora#4710](https://github.com/hashicorp/pandora/pull/4710).

Thanks for your response @wuxu92. That's fine, but we just want to be able to know about these changes to our service's directory. We want to know what changes and when happen to Device Registry here. Is there any reason Hashicorp would prefer we don't update the CODEOWNERS to do this?

stephybun:
Hey @mryanlo, due to the way the [CODEOWNERS file currently works](https://github.com/orgs/community/discussions/23042) these additions won't achieve the desired outcome of notifying your team of changes in the `deviceregistry` SDK since the GH users referenced in the file need to have write access on the repo.

image

Since this can't work without write access we're not able to accept this change.

PR #1148, opened 2025-01-24
fix: Use url.JoinPath() to more flexibly handle extraneous slashes

bmoylan:
In a custom cloud environment, we configure terraform-provider-azurerm with a `metadata_host` whose response to `GET /metadata/endpoints?api-version=2022-09-01` includes a trailing slash in the `loginEndpoint` field. The current `Sprintf` approach assumes well formed (no slashes) endpoints. The provider fails to get OIDC credentials in this environment due to the double slash at the beginning of the path returned by tokenEndpoint().

In order to be more robust to this subtly unepected input, the provider can use [url.JoinPath](https://pkg.go.dev/net/url#JoinPath) to clean missing and double slashes.


bmoylan:
Hi @manicminer it looks like you've worked on a lot of the surrounding code here. Would you be able to review this change? Thanks in advance!

PR #1218, opened 2025-07-04
Fix disk grant access

tianyiliuMSFT:
## Community Note

## Description

1. Access URI of the disk should be returned from `GrantAccessThenPoll`, so the return value of this function is changed.
2. `AccessUri` in `GrantAccessOperationResponse` doesn't match the actual one, which leads to unmarshal failure. Added and changed the fields of the struct.

- [x] Bug Fix
- [ ] New Feature
- [ ] Enhancement
- [ ] Breaking Change

## Related Issue(s)
Fixes #0000

## Rollback Plan

> [!NOTE] 

### DIFF
--- a/resource-manager/compute/2022-03-02/disks/method_grantaccess.go
+++ b/resource-manager/compute/2022-03-02/disks/method_grantaccess.go
@@ -48,6 +48,9 @@ func (c DisksClient) GrantAccess(ctx context.Context, id commonids.ManagedDiskId
 	if resp != nil {
 		result.OData = resp.OData
 		result.HttpResponse = resp.Response
+
+		var model AccessUri
+		result.Model = &model
 	}
 	if err != nil {
 		return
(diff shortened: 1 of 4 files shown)


jackofallops:
Hi @tianyiliuMSFT - unfortunately these files are generated and this change will be overwritten.  The LRO should not return the URI of the created resource afaik, and a subsequent get on the ID can be used to retrieve the necessary data. 

Since this PR would simply be overwritten by the next ingestion of the API specs, I'm going to have to close it out.

arsdragonfly:
> Hi @tianyiliuMSFT - unfortunately these files are generated and this change will be overwritten.  The LRO should not return the URI of the created resource afaik, and a subsequent get on the ID can be used to retrieve the necessary data. 
> 
> Since this PR would simply be overwritten by the next ingestion of the API specs, I'm going to have to close it out. 

@jackofallops This change is important for maintaining the VHD generation feature of Packer plugin for Azure. The existing method of generating VHD is via Unmanaged Disks, which is unsupported on newer SKUs and is [completely going out of support this September.](https://learn.microsoft.com/en-us/azure/virtual-machines/unmanaged-disks-deprecation)
Unfortunately the ephemeral SAS url can only be retrieved from polling the LRO. The URL isn't available via GET.
If possible, would you guide us on how we can make changes to your SDK generator to support such LRO result gathering?

PR #1221, opened 2025-07-15
`sdk` - rework `LRO` and `ProvisioningState` pollers to match ARM spec for terminal statuses

jackofallops:
This PR reworks how the SDK deals with Long Running Operation polling by removing custom "Terminal" statuses from consideration. ARM allows arbitrary values in the `Status` fields, including custom values supplied by users. Since it is not possible to know all possible values in advance, this PR delegates the success/failure of an operation to the service returning one of the ARM specified Terminal Statuses in the LRO response. Any other values are considered "InProgress" for the purposes of Operation Status.

This PR includes the changes in #1124 and supersedes it.

### DIFF
--- a/sdk/client/resourcemanager/poller_lro.go
+++ b/sdk/client/resourcemanager/poller_lro.go
@@ -238,3 +238,11 @@ const (
 	statusInProgress status = "InProgress"
 	statusSucceeded  status = "Succeeded"
 )
+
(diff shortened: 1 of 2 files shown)


mbfrahry (approved):
I just have one concern that I've left a comment on but if that isn't an issue, this looks good!

PR #1228, opened 2025-07-31
`ProvisioningStatePoller` - reverting removal of workaround for faux LRO early exit

jackofallops:
renames shadowed var too


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #1287, opened 2026-01-28
add custom headers for MSI authorizer

ziyeqf:
## Community Note

## Description

In specific scenarios (Azure Container App or Service Fabric), the requests need additional headers.  Reference: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/22fd2ee68abf6e6e926fa09387c078db25dad7e0/apps/managedidentity/managedidentity.go#L292-L296

- [ ] Bug Fix
- [ ] New Feature
- [x] Enhancement
- [ ] Breaking Change

## Related Issue(s)
Fixes #0000

a related issue on terraform repo: https://github.com/hashicorp/terraform/issues/37268

## Rollback Plan

> [!NOTE] 

### DIFF
--- a/sdk/auth/managed_identity_authorizer.go
+++ b/sdk/auth/managed_identity_authorizer.go
@@ -32,6 +32,10 @@ type ManagedIdentityAuthorizerOptions struct {
 	// CustomManagedIdentityAPIVersion is an optional API version to use when requesting a token.
 	// This is required when using an endpoint that does not support the default API version such as Azure Container Apps.
 	CustomManagedIdentityAPIVersion string
+
+	// CustomManagedIdentityHeaders are optional HTTP headers to use when requesting a token
+	// This is required when requesting from Azure Container App or Service Fabric.
+	CustomManagedIdentityHeaders map[string][]string
 }
 
 // NewManagedIdentityAuthorizer returns an authorizer using a Managed Identity for authentication.
@@ -40,7 +44,7 @@ func NewManagedIdentityAuthorizer(ctx context.Context, options ManagedIdentityAu
 	if err != nil {
 		return nil, fmt.Errorf("determining resource for api %q: %+v", options.Api.Name(), err)
 	}
-	conf, err := newManagedIdentityConfig(*resource, options.ClientId, options.CustomManagedIdentityEndpoint, options.CustomManagedIdentityAPIVersion)
+	conf, err := newManagedIdentityConfig(*resource, options.ClientId, options.CustomManagedIdentityEndpoint, options.CustomManagedIdentityAPIVersion, options.CustomManagedIdentityHeaders)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


wuxu92 (commented):
Thanks @ziyeqf for the feature! it looks good to me overall, except one I think we can simplify the header process.

sdk/auth/managed_identity_authorizer.go
```diff
@@ -165,14 +173,22 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
...
+
+	if headers != nil {
+		if _, ok := headers["Metadata"]; !ok {
+			headers["Metadata"] = []string{"true"}
+		}
+		req.Header = headers
+	} else {
+		req.Header = http.Header{
+			"Metadata": []string{"true"},
+		}
```
wuxu92:
```suggestion
	req.Header = http.Header{
		"Metadata": []string{"true"},
	}
	
	for k, v := range headers {
		req.Header[k] = v
	}
```

sdk/auth/managed_identity_authorizer.go
```diff
@@ -165,16 +174,19 @@ func (c *managedIdentityConfig) TokenSource(_ context.Context) (Authorizer, erro
...
 	req, err = http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
 	if err != nil {
 		return
 	}
+
 	req.Header = http.Header{
 		"Metadata": []string{"true"},
 	}
 
+	maps.Copy(req.Header, headers)
```
jackofallops: Given how sensitive some APIs are, maybe we guard against accidental input and force canonical naming?
```suggestion
	for k, v := range headers {
		req.Header[http.CanonicalHeaderKey(k)] = v
	}
```

jackofallops (changes requested):
Thanks @ziyeqf - One minor change to take a look at.  The referenced lib suggests that the env var is always the same `IDENTITY_HEADER` -  is the plan to have the provider always, or conditionally, look for this in the `providerConfigure()` call?

ziyeqf:
Hi @jackofallops, yes I think so, the main pain point is to let the sdk can read what needed from env vars.. Possibly a compromise way is adding a new feature flag let the use specify the name of the environment variable. WDYT?

jackofallops:
> Hi @jackofallops, yes I think so, the main pain point is to let the sdk can read what needed from env vars.. Possibly a compromise way is adding a new feature flag let the use specify the name of the environment variable. WDYT?

We should probably pick up this discussion on the Provider when the PR is opened, but as a starting point I think we should probably default to reading and using the value of that env var if it's present, and defaulting to `nil` otherwise (this way it's a no-op for users in the affected environments, and typically the env var won't be set outside of the "non-standard" environments?).  There may be argument / case for a configurable option if we know of other names that var might go by, or if there's genuine value in being able to make it configurable. 

For now I'll approve and merge this one.

jackofallops (approved):
Thanks @ziyeqf for the changes, LGTM now 👍

PR #1324, opened 2026-04-29
`pollers` - keep polling until deadline when caller's context is canceled

pszypowicz:
## Community Note

## Description

When the caller's context is canceled (for example, SIGINT during `terraform apply`), `PollUntilDone` returns `context.Canceled` immediately. The Azure operation that was already accepted server-side continues to completion, leaving an orphaned resource that callers have to recover via manual `terraform import`.

This change runs the polling goroutine on a context decoupled from the caller's via `context.WithoutCancel`, preserving the caller's deadline. On `context.Canceled`, polling continues until either the operation completes or the deadline elapses. `context.DeadlineExceeded` returns immediately - the deadline is the caller's contract.

The deadline is therefore the natural upper bound on the grace window: zero new API surface, no configuration, and no behavior change when the caller is not actively canceling.

### Scope

This change addresses the **polling layer only**. Resources whose Create function does post-poll work using the same `ctx` (a Read-after-Create at the wrapper level, data-plane availability waits for storage accounts, kubelet identity probes for AKS, etc.) will still produce a `tainted` resource on cancel - because that follow-up work also fails with `context canceled` once the poll returns.

That is still a meaningful UX improvement over the baseline: the resource ID lands in state, recoverable with `terraform untaint` (or destroyed-and-recreated on the next `terraform apply`), instead of being orphaned and requiring manual `terraform import`. End-to-end validation against a real Azure subscription confirmed this:

| Scenario | Without this change | With this change |
|---|---|---|
| Create, SIGINT mid-poll | resource orphaned in Azure, missing from state | resource in state (tainted), exists in Azure |
| Create, SIGINT + SIGTERM | resource orphaned (escape hatch preserved) | resource orphaned (escape hatch preserved) |
| 10x parallelism=2, SIGINT after 2 in-flight | only the 2 in-flight orphan; the other 8 not started | the 2 in-flight land in state (tainted); the other 8 not started |

A follow-up at the provider layer is needed to remove the `tainted` marker; that is out of scope for the SDK.

### Test plan

- 6 new unit tests in `sdk/client/pollers/poller_test.go`:
  - `TestPoller_GraceOnCancel_PollCompletesBeforeDeadline`
  - `TestPoller_GraceOnCancel_DeadlineCapsGrace`
  - `TestPoller_GraceOnCancel_DeadlineExceededReturnsImmediately`
  - `TestPoller_GraceOnCancel_PollFailsDuringGrace`
  - `TestPoller_GraceOnCancel_DroppedConnectionsThenSucceeds`
  - `TestPoller_GraceOnCancel_RepeatedCancelIsIdempotent`
- `go test -race -count=10 ./client/pollers/...` passes locally.
- Whole `sdk` module green under `-race`.

- [ ] Bug Fix
- [ ] New Feature
- [x] Enhancement
- [ ] Breaking Change

## Related Issue(s)

Refs hashicorp/terraform-provider-azurerm#32230 - azurerm-side request, with the end-to-end validation results above.
Refs hashicorp/terraform-provider-google#9874 - same problem on a sibling provider, open since 2021; the reporter explicitly asked for the design implemented here.
Refs hashicorp/terraform-provider-azurerm#28513 - closest prior provider-side attempt; closed pending an internal design decision.

Out of scope: poll-time API failures (e.g. hashicorp/terraform-provider-azurerm#28511 image-pull failure on `azurerm_container_app`). That is a separate SetID-timing question, not addressed here.

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the SDK.

No changes to security controls. The grace window is bounded by the caller's existing deadline, which is already an accepted contract; no new surface is exposed for prolonged operations.

> [!NOTE]

### DIFF
--- a/sdk/client/pollers/poller.go
+++ b/sdk/client/pollers/poller.go
@@ -126,6 +126,18 @@ func (p *Poller) PollUntilDone(ctx context.Context) error {
 		return fmt.Errorf("internal-error: `ctx` should have a deadline")
 	}
 
+	// pollCtx drives the inner polling goroutine. It is decoupled from ctx so that
+	// when the caller cancels (e.g. SIGINT during `terraform apply`) we can keep
+	// polling and let the in-flight Azure operation complete - long enough for the
+	// resource ID to be persisted into state on the next layer up. pollCtx inherits
+	// the caller's deadline, so the resource's existing `timeouts.create` value
+	// naturally caps how long polling continues. The caller already declared that
+	// deadline as their patience ceiling; honoring it post-cancel is consistent
+	// with that declaration.
+	deadline, _ := ctx.Deadline()
+	pollCtx, cancelPoll := context.WithDeadline(context.WithoutCancel(ctx), deadline)
+	defer cancelPoll()
+
 	var wait sync.WaitGroup
 	wait.Add(1)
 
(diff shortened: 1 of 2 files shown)


sreallymatt:
Hi @pszypowicz - this is a behavioural change that we cannot force on all users of the provider, while this may make sense for your use-case, we cannot assume that is the case for everyone.

We're working on adding [a feature to the provider](https://github.com/hashicorp/terraform-provider-azurerm/pull/32318) that will allow users to opt in to setting the resource ID after the initial request to Azure succeeds, before polling for completion, which will allow for an early cancellation to still track the remote resource(s).

PR #1336, opened 2026-05-20
Auto PR: Regenerating the Go SDK (60b392664e5d39fa4f80e63869852a7881f9bb81)

hc-github-team-tf-azure:
Regenerating the SDK based on the latest changes


sreallymatt (approved):
Replaced in the `azurerm` provider and confirmed there are no compilation errors. This PR adds the `CallbackThenPoll` method to all LROs and updates the `ThenPoll` methods to call it, passing `nil` for the callback func parameter which preserves the behaviour of existing `ThenPoll` functions.

PR #1337, opened 2026-05-29
restore defaultRetryPolicy behaviour when we're on AzureRM 5.0

jackofallops:


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #1340, opened 2026-06-02
`actions` - update `pr-unit-tests.yaml` to RAM disk and matrix testing for speed

jackofallops:
Note: ci targets do run fmt since if the PR's open that should have already been done! (and was insanely "expensive" on execution time 💸 )


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #1343, opened 2026-06-03
`gha` - reduce release action bottleneck

jackofallops:
Note: We could arguably remove the testing step entirely as the `pr-unit-tests.yaml` covers this before we merge anything anyway? I've left it in as this should already save us around 80% on time waiting.

also added a collected `test-ci` make target for developer convenience.


catriona-m (approved):
Thanks @jackofallops LGTM!

PR #1365, opened 2026-08-31
pollers - allow `PollerFromResponse` to return poller based on async-operation URL for Delete requests

sreallymatt:
## Community Note

## Description

Removes a single condition to allow `Delete` operations to poll on the returned async operation URL rather than just polling for a 404 on the resource URL.

Reasoning behind this change:

Certain resources in Azure may accept an initial `Delete` request, but fail later on due to some condition. Because the current logic just checks for a 404, the operation keeps polling until the context expires. At which point it surfaces only a `context deadline exceeded` error to the user rather than an actionable error message.

`azurerm_postgresql_flexible_server_active_directory_administrator` is an example of this, if you attempt a delete while certain objects still depend on it (e.g. setting this admin object as the owner of a table in a database), Azure accepts the initial delete request with a `202` response but performing a `GET` on the returned async operation url (from `azure-asyncoperation` header) returns the following error:

```
Status: "AadAuthPrincipalDropFailed"
│ Code: ""
│ Message: "Failed to drop Microsoft Entra principal. Reason: '2BP01: role \"acctest-fs-0831-grp\" cannot be dropped because some objects depend on it\n\nDETAIL: Detail redacted as it may contain sensitive data. Specify 'Include Error Detail' in the connection string to include this information.'."
```

which is never surfaced to the user, and the resource proceeds to poll until the 30 min timeout expires.

- [ ] Bug Fix
- [ ] New Feature
- [x] Enhancement
- [ ] Breaking Change

## Rollback Plan

> [!NOTE] 

### DIFF
--- a/sdk/client/resourcemanager/poller.go
+++ b/sdk/client/resourcemanager/poller.go
@@ -25,7 +25,7 @@ func PollerFromResponse(response *client.Response, client *Client) (poller polle
 	methodIsDelete := strings.EqualFold(response.Request.Method, "DELETE")
 	lroPollingUri := pollingUriForLongRunningOperation(response)
 	lroIsSelfReference := isLROSelfReference(lroPollingUri, originalRequestUri)
-	if isLroStatus && lroPollingUri != "" && !methodIsDelete && !lroIsSelfReference {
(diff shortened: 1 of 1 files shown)


mstroob (approved):
LGTM 🇧🇸

